PR #23, opened 2023-02-16
Add code coverage statistics

ConnorStoneAstro:
Integrates with codecov to display code coverage statistics

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -42,13 +42,13 @@ jobs:
       shell: bash
     - name: Install Caustic
       run: |
-        cd $GITHUB_WORKSPACE/caustic
+        cd $GITHUB_WORKSPACE
         pip install -e .
         pip show caustic
       shell: bash
     - name: Test with pytest
       run: |
-        cd $GITHUB_WORKSPACE/caustic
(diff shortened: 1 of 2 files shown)


ConnorStoneAstro:
Everything is working, ready for merge!

adam-coogan:
Very cool, thanks for your work on this! Two minor points:

- Is it necessary to have separate actions for the coverage check and tests? They're both calling pytest, so can they be combined somehow?
- Also, have you checked the badge displays correctly? It appears gray on [the PR branch](https://github.com/Ciela-Institute/caustic/tree/addcoverage), but I guess the workflow is set up to run on the `main` branch.

ConnorStoneAstro:
You could combine the tests and checks, but the checks run on Linux MacOS and windows, while the coverage should really only be run once. I could probably combine them but I don't think it's worth the effort. Kind of nice to have them separate since they are different things. Also the code coverage will fail if the coverage decreases, and we don't want that to get misconstrued as the tests failing. 

As for the badge, it only reads from the main branch, so it will only change once the coverage code has been merged. There's probably a way to check ahead of time, but for AutoProf I just merged it and everything worked right away.

adam-coogan (approved):
Looks great!

PR #24, opened 2023-02-21
debug indexing error in multiplane.py

ConnorStoneAstro:
Indexing erroneously assumed the lenses were sorted by redshift, now they are accessed in the correct order.


ConnorStoneAstro:
This has been resolved elsewhere. Closing because it is no longer relevant.

Issue #30, opened 2023-04-21
torchinterp1d and KeOps are incompatible with vmap


adam-coogan:
`vmap` is not compatible with `torchinterp1d` or `keops`. Both use `torch.autograd.Function`, which is not compatible with `vmap` by default. This will be a problem if you want to e.g. `vmap` over a bunch of `NFW`s for simulating or computing magnifications, since they may use `torchinterp1d` via `FlatLambdaCDM`. [This is fixable](https://pytorch.org/docs/stable/notes/extending.func.html#torch-vmap-support), but requires modifying `torchinterp1d` and `keops`. I've opened issues in [both](https://github.com/aliutkus/torchinterp1d/issues/20) projects' [repos](https://github.com/getkeops/keops/issues/303).

This isn't really our responsibility to fix, but I wanted to document this since it's good to be aware of this restriction.

ConnorStoneAstro:
We no longer depend on torchinterp1d and only have limited compatibility with keops. It would be good to open another issue specifically for keops compatibility.

PR #36, opened 2023-05-06
Update Kappa Grid fft and dtype behavior

### LINKED ISSUE #35: Kappa grid complex dtype
We should not change the dtype of complex variable so freely, it can break the module. Right now, the .to method looks like this

```
    def to(
        self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None
    ):
        super().to(device, dtype)
        self.Psi_kernel = self.Psi_kernel.to(device=device, dtype=dtype)
        self.ax_kernel = self.ax_kernel.to(device=device, dtype=dtype)
        self.ay_kernel = self.ay_kernel.to(device=device, dtype=dtype)
        if self.Psi_kernel_tilde is not None:
            self.Psi_kernel_tilde = self.Psi_kernel_tilde.to(device=device, dtype=dtype) # Dont do this
        if self.ax_kernel_tilde is not None:
            self.ax_kernel_tilde = self.ax_kernel_tilde.to(device=device, dtype=dtype)
        if self.ay_kernel_tilde is not None:
            self.ay_kernel_tilde = self.ay_kernel_tilde.to(device=device, dtype=dtype)

```

Instead, we should not allow the user to change the dtype of those specific variable, like this

```
    def to(
        self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None
    ):
        super().to(device, dtype)
        self.Psi_kernel = self.Psi_kernel.to(device=device, dtype=dtype)
        self.ax_kernel = self.ax_kernel.to(device=device, dtype=dtype)
        self.ay_kernel = self.ay_kernel.to(device=device, dtype=dtype)
        if self.Psi_kernel_tilde is not None:
            self.Psi_kernel_tilde = self.Psi_kernel_tilde.to(device=device)
        if self.ax_kernel_tilde is not None:
            self.ax_kernel_tilde = self.ax_kernel_tilde.to(device=device)
        if self.ay_kernel_tilde is not None:
            self.ay_kernel_tilde = self.ay_kernel_tilde.to(device=device)

```

ConnorStoneAstro:
Kappa Grid is updated with the following:
- dtype no longer enforced for frequency space kernels
- docstrings added
- fft changed to rfft since kernels are real valued
- enforced shapes for fft and ifft

This PR closes #35

### DIFF
--- a/src/caustic/lenses/kappa_grid.py
+++ b/src/caustic/lenses/kappa_grid.py
@@ -28,12 +28,32 @@ class KappaGrid(ThinLens):
         mode: str = "fft",
         use_next_fast_len: bool = True,
     ):
+        """Strong lensing with user provided kappa map
+
+        KappaGrid is a class for strong gravitational lensing with a
+        user-provided kappa map. It inherits from the ThinLens class.
+        This class enables the computation of deflection angles and
+        lensing potential by applying the user-provided kappa map to a
+        grid using either Fast Fourier Transform (FFT) or a 2D
+        convolution.
+
+        Attributes:
+            name (str): The name of the KappaGrid object.
+            fov (float): The field of view in arcseconds.
+            n_pix (int): The number of pixels on each side of the grid.
+            cosmology (Cosmology): An instance of the cosmological parameters.
+            z_l (Optional[Tensor]): The redshift of the lens.
+            thx0 (Optional[Tensor]): The x-coordinate of the center of the grid.
+            thy0 (Optional[Tensor]): The y-coordinate of the center of the grid.
+            kappa_map (Optional[Tensor]): A 2D tensor representing the kappa map.
+            kappa_map_shape (Optional[tuple[int, ...]]): The shape of the kappa map.
+            mode (str, optional): The convolution mode for calculating deflection angles and lensing potential.
+                It can be either "fft" (Fast Fourier Transform) or "conv2d" (2D convolution). Default is "fft".
+            use_next_fast_len (bool, optional): If True, adds additional padding to speed up the FFT by calling
+                `scipy.fft.next_fast_len`. The speed boost can be substantial when `n_pix` is prime. Default is True.
+
         """
-        Args:
-            use_next_fast_len: if true, add additional padding to speed up the FFT
-                by calling `scipy.fft.next_fast_len <https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.next_fast_len.html#scipy.fft.next_fast_len>`_.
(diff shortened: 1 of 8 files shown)


ConnorStoneAstro:
Now torchinterp1d has been eliminated, this is related to #30

src/caustic/lenses/kappa_grid.py
```diff
@@ -28,12 +27,32 @@ def __init__(
...
+            cosmology (Cosmology): An instance of the cosmological parameters.
+            z_l (Optional[Tensor]): The redshift of the lens.
+            thx0 (Optional[Tensor]): The x-coordinate of the center of the grid.
+            thy0 (Optional[Tensor]): The y-coordinate of the center of the grid.
+            kappa_map (Optional[Tensor]): A 2D tensor representing the kappa map.
+            kappa_map_shape (Optional[tuple[int, ...]]): The shape of the kappa map.
+            mode (str, optional): The convolution mode for calculating deflection angles and lensing potential.
+                It can be either "fft" (Fast Fourier Transform) or "conv2d" (2D convolution). Default is "fft".
+            use_next_fast_len (bool, optional): If True, adds additional padding to speed up the FFT by calling
+                `scipy.fft.next_fast_len`. The speed boost can be substantial when `n_pix` is prime. Default is True.
```
AlexandreAdam, line 52: I believe the speedup happens when n_pix is a multiple of some factors that make FFT efficient. I copy here scipy's doc string so that we can make ours reflect theirs better.

SciPy’s FFT algorithms gain their speed by a recursive divide and conquer strategy. This relies on efficient functions for small prime factors of the input length. Thus, the transforms are fastest when using composites of the prime factors handled by the fft implementation. If there are efficient functions for all radices <= n, then the result will be a number x >= target with only prime factors < n. (Also known as n-smooth numbers)

AlexandreAdam (approved):

AlexandreAdam:
Before merging this, we should also solve the issue in utils/interp2d found, namely that the line
```
  if padding_mode == "zeros":  # else padding_mode == "extrapolate"
      result[idxs_out_of_bounds] = 0.0
```
can cause torch to complain that 0.0 is not on the correct device when using GPU. Instead, we found that
```
  if padding_mode == "zeros":  # else padding_mode == "extrapolate"
      result[idxs_out_of_bounds] = torch.zeros_like(result[idxs_out_of_bounds])
```
works in any context (cpu or gpu).

ConnorStoneAstro:
Looks like I have the verification working. I'm not going to rebase all my old commits, I'll just override the signing protection this time.

PR #37, opened 2023-05-09
Simulator class

ConnorStoneAstro:
Adds subclass of Parametrized which defines a simulator object. This defines a behaviour for the `__call__` method to handle parameter packing then pass to a user defined forward method.

### DIFF
--- a/src/caustic/parametrized.py
+++ b/src/caustic/parametrized.py
@@ -12,7 +12,6 @@ from .parameter import Parameter
 
 __all__ = ("Parametrized", "Simulator")
 
-
 class Parametrized:
     """
     Represents a class with Param and Parametrized attributes.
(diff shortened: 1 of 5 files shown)


ConnorStoneAstro (commented):
This one is pretty basic

src/caustic/parametrized.py
```diff
@@ -401,7 +400,21 @@ def add_params(p: Parametrized, dot):
 
         return dot
 
+class Simulator(Parametrized):
```
AlexandreAdam, line 403: This should be in a separate file called simulator.py. Even though it is a very small wrapper, we might add stuff in the future.

AlexandreAdam (changes requested):

src/caustic/parametrized.py
ConnorStoneAstro: Ok, now its in a separate file

AlexandreAdam (approved):

PR #43, opened 2023-06-08
Renaming almost everything in caustic to be more explicit

ConnorStoneAstro:
Almost everything renamed and restructured

- Greek letters `alpha`, `kappa`, `psi` renamed to `deflection_angle`, `convergence`, `potential`
- deflection_angle made more explicit with `reduced_deflection_angle` and `physical_deflection_angle`
- ThickLens lensing made more explicit, base deflection angle calculator is pure `raytrace`, now `physical_deflection_angle` throws an error, `reduced_deflection_angle` throws a warning and points to `effective_reduced_deflection_angle`
- project moved from `src` directory to simpler `caustic` directory
- cosmology made more explicit with `dist` to `distance` and `Sigma_cr` to `critical_surface_density`
- Pixelated source and lens made more explicit with source now called `Pixelated` and lens now called `PixelatedConvergence` leaving open for future work: `PixelatedPotential` and other variants
- Sersic parameters renamed to match literature with `index` going to `n` also now `q` is a true axis ratio which stretches only the minor axis and not an alternative definition which stretches both axes


AlexandreAdam:
I think we should be avoiding capital letter for variables. Right now, we redefined `x` to `P`.  I agrre that small 'p' is not a good option, so we need to agree on something different. 

I do like `x`, since this is always how I write inverse problems. I suppose other bayesian use `theta`, but both seem to be used in some modules. 

As an alternative, we could be more explicit and call it `params`. I think everybody should pitch in since this is the most important variable in the code.

AlexandreAdam:
Also while we are at it, we should change the docstring for the variable `P` from `additional parameters` to `dynamical parameters` or something that puts this varibale more in the spot light.

caustic/lenses/epl.py
```diff
@@ -131,45 +131,45 @@ def _r_omega(self, z, t, q):
...
         Compute the convergence of the lens, which describes the local density of the lens.
 
         Args:
-            thx (Tensor): X coordinates in the lens plane.
-            thy (Tensor): Y coordinates in the lens plane.
+            x (Tensor): X coordinates in the lens plane.
+            y (Tensor): Y coordinates in the lens plane.
             z_s (Tensor): Source redshifts.
-            x (Optional[dict[str, Any]]): Additional parameters for the lens model.
+            x (Packed): Additional parameters for the lens model.
```
AlexandreAdam, line 166: Shouldn't that be `P`?

caustic/lenses/multiplane.py
```diff
@@ -28,45 +28,47 @@ def __init__(self, name: str, cosmology: Cosmology, lenses: list[ThinLens]):
...
             self.add_parametrized(lens)
 
-    def get_z_ls(self, x: Optional[dict[str, Any]]) -> list[Tensor]:
+    def get_z_ls(self, P: "Packed") -> list[Tensor]:
         """
         Get the redshifts of each lens in the multiplane.
 
         Args:
-            x (Optional[dict[str, Any]]): Additional parameters.
+            x (Packed): Additional parameters.
```
AlexandreAdam, line 36: Same here, should be `P`

caustic/lenses/nfw.py
```diff
@@ -198,115 +198,117 @@ def _h(cls, x: Tensor) -> Tensor:
...
-
-    def kappa(
-        self, thx: Tensor, thy: Tensor, z_s: Tensor, x: Optional[dict[str, Any]] = None
+        ax = deflection_angle * x / th
+        ay = deflection_angle * y / th
+        d_s = self.cosmology.angular_diameter_distance(z_s, P)
+        d_ls = self.cosmology.angular_diameter_distance_z1z2(z_l, z_s, P)
+        return ax * d_ls / d_s, ay * d_ls / d_s
+
+    # def deflection_angle(
```
AlexandreAdam, line 242: Why is this commented?

AlexandreAdam (changes requested):
I requested some changes. Feel free to resolve them with the changes that you find appropriate. Good work on this PR!

andreasfilipp:
I'd also change `P` to `params/parameters` or `x`. Besides that I do not have anything to add to that PR

caustic/lenses/epl.py
ConnorStoneAstro: Yup, fixed

caustic/lenses/nfw.py
ConnorStoneAstro: I was making sure I had the reduced/physical definitions right, it's good now

AlexandreAdam (approved):
Looks good to me. Thanks Connor!

Issue #44, opened 2023-06-26
Making the 'name' requirement optional


andreasfilipp:
For an easy usage it might be easier to use if the 'name' requirement for mass and light profiles remains optional and they get automatically unique names.

AlexandreAdam:
I second that opinion, it shouldn't be too hard to implement this both at the module and simulator level to avoid name collision. Basically, we need to give the Simulator class a procedure to recursively update the name of the modules that share the same name.

Issue #48, opened 2023-06-30
Implement "to" method for all objects


ConnorStoneAstro:
A recursively applied to method should put every pytorch object onto the correct dtype and device. Should maybe also be a method for "packed"

AlexandreAdam:
Looking inside the `Parametrized` module, it seems the `to` method is already recursive. We should probably just add test to make sure the `to` method behaves as we expect. 

The [Parametrized update](https://github.com/Ciela-Institute/caustic/pull/50) already has some changes to the module and added test. We could add those there.

PR #52, opened 2023-07-02
Parameterized name update

AlexandreAdam:
# Parametrized name update / refactoring

**Note**: This PR is stacked on top of [Parametrized update](https://github.com/Ciela-Institute/caustic/pull/50), which should be resolved first. 

This PR is related to [Issue 44](https://github.com/Ciela-Institute/caustic/issues/44).

## Content
- Refactoring of `Parametrized` to support default name in each module and handling of name collision
- Improvement of NestedNamespaceDict, which in turns has enabled a refactoring of DAG related methods in `Parametrized`, making the methods more intuitive and the code less clunky. 
- A suite of tests are added 
- Module are updated so that the `name` is now optional and default to the class name. The position of the `name` argument has moved to the last position in each `__init__` call. This will break some of the implementation of caustic if they used positional argument.


AlexandreAdam:
Now that I think about it, I should probably rewrite some of these tests to be smaller and more specific. Always helps with debugging. 

Also, I think we should update the Parameter module so that it also supports renaming.

andreasfilipp:
> Now that I think about it, I should probably rewrite some of these tests to be smaller and more specific. Always helps with debugging.
> 
> Also, I think we should update the Parameter module so that it also supports renaming.

I agree - the Parameter module should probably support renaming (If it's easy implementable). Besides I do not have other comments.

andreasfilipp (approved):

caustic/parametrized.py
```diff
@@ -282,59 +169,55 @@ def pack(
...
         elif isinstance(x, list) or isinstance(x, tuple):
             n_passed = len(x)
-            n_expected = (
-                sum([desc.n_dynamic for desc in self._descendants.values()])
-                + self.n_dynamic
-            )
-            if n_passed != n_expected:
+            n_dynamic_params = len(self.params.dynamic.flatten())
+            n_dynamic_modules = len(self.dynamic_modules)
+            if n_passed not in [n_dynamic_modules, n_dynamic_params]:
```
ConnorStoneAstro, line 183: I would change this clause to just an else at the end of this statement. ie:
```
if n_passed == n_dynamic_params:
   ...
elif n_passed == n_dynamic_modules:
    ...
else:
    raise ValueError(...)
    ```

File: caustic/parametrized.py [hunk 2]
```diff
@@ -282,59 +169,55 @@ def pack(
...
-            ]
+            missing_names = [name for name in self.params.dynamic.keys() if name not in x]
             if len(missing_names) > 0:
                 raise ValueError(f"missing x keys for {missing_names}")
 
             # TODO: check structure!
-
             return Packed(x)
+
         elif isinstance(x, list) or isinstance(x, tuple):
```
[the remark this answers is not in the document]
ConnorStoneAstro: Alternatively you could accept any iterable type by just checking `hasattr` for `__iter__`. But then you would need to check for anything specific (Tensor) first, before accepting a general iterable.

caustic/parametrized.py
```diff
@@ -429,72 +312,59 @@ def __getattribute__(self, key):
...
-
-        Returns:
-            OrderedDict[str, Parameter]: An ordered dictionary of static parameters.
-        """
-        return self.params["static"]
+    def params(self) -> NestedNamespaceDict:
+        static = NestedNamespaceDict()
+        dynamic = NestedNamespaceDict()
+        def _get_params(module):
+            if module._childs != {}:
```
ConnorStoneAstro, line 330: maybe `if len(module._childs) > 0`, its more general

caustic/parametrized.py
```diff
@@ -429,72 +312,59 @@ def __getattribute__(self, key):
...
 
     def __str__(self) -> str:
-        """
-        Returns a string description of the object.
-
-        Returns:
-            str: A string description of the object.
-        """
-        static, dynamic = itemgetter("static", "dynamic")(self.params)
+        static, dynamic = itemgetter("static", "dynamic")(self.module_params)
```
ConnorStoneAstro, line 358: Couldnt this just be:
`static = self.module_params["static"]`
`dynamic = self.module_params["dynamic"]`
its easier to read

caustic/parametrized.py
```diff
@@ -520,16 +390,14 @@ def get_graph(
...
-
         import graphviz
 
         def add_component(p: Parametrized, dot):
             dot.attr("node", style="solid", color="black", shape="ellipse")
             dot.node(p.name, f"{p.__class__.__name__}('{p.name}')")
 
         def add_params(p: Parametrized, dot):
-            static, dynamic = itemgetter("static", "dynamic")(p.params)
+            static, dynamic = itemgetter("static", "dynamic")(p.module_params)
```
ConnorStoneAstro, line 400: same here, could be two simple lines

test/test_namespace_dict.py
```diff
@@ -0,0 +1,98 @@
...
+    nested_namespace["bar.qux.corge"] = 98
+    print(nested_namespace.flatten())
+    assert len(list(nested_namespace.flatten().keys())) == 4
+
+
+def test_nested_namespace_dict_collapse():
+    nested_namespace = NestedNamespaceDict()
+    nested_namespace["foo"] = 1
+    nested_namespace["bar"] = {"baz": 2}
+    nested_namespace["bar.qux"] = {"quux": 42}
```
ConnorStoneAstro, line 63: Should add an assignment here to explicitly test collision/overwriting behavior

test/test_parametrized.py
```diff
@@ -0,0 +1,192 @@
...
+
+    sim = Sim()
+
+    cosmo_params = [0.7]
+    lens_params = [0.5, 0., 0., 0.8, 0.5, 1.2, 1.1]
+    source_params = [0.0, 0.0, 0.8, 0.0, 1., 0.2, 10.]
+    
+    # test list input
+    x = [torch.tensor(_x) for _x in cosmo_params + lens_params + source_params]
+    sim.forward(sim.pack(x))
```
ConnorStoneAstro, line 66: Simulator object already calls pack on inputs so you can just do `sim(x)`

Also pack accepts list inputs, can't you just do: `sim(cosmo_params + lens_params + source_params)`?

test/test_parametrized.py
```diff
@@ -0,0 +1,192 @@
...
+    sim(x)
```
ConnorStoneAstro, line 67: here you have `sim(x)` so why also have `sim.forward(sim.pack(x))`?

test/test_parametrized.py
```diff
@@ -0,0 +1,192 @@
...
+
+    sim = Sim()
+
+    lens_params = [0.5, 0., 0., 0.8, 0.5, 1.2, 1.1]
+    source_params = [0.0, 0.0, 0.8, 0.0, 1., 0.2, 10.]
+    
+    # test list input
+    x = [torch.tensor(_x) for _x in lens_params + source_params]
+    sim.forward(sim.pack(x))
+    sim(x)
```
ConnorStoneAstro, line 113: maybe check type, size etc of the output for one of the sim calls?

ConnorStoneAstro (changes requested):
Very nice work! I have some comments to look through that are really minor stuff.

caustic/namespace_dict.py
```diff
@@ -0,0 +1,178 @@
...
+        #==============================
+        # Flatten and collapse method
+        #==============================
+        print(nested_dict.flatten())
+        # Output:
+        # {'foo': 'Hello', 'foo.bar.baz': 'World', 'foo.bar.qux': 42}
+        
+        print(nested_dict.collapse()
+        # Output:
+        # {'foo': 'Hello', 'baz': 'World', 'qux': 42}
```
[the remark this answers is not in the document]
AlexandreAdam: The purpose of collapse is precisely to create collision, but is not used yet in the code. I was thinking it would be useful if we have DAG structure that enforces unique name for each of its module. A collision, in such a case, would allow us to resolve a node with multiple edges as a unique object.

caustic/parametrized.py
```diff
@@ -24,124 +25,84 @@ class Parametrized:
...
+        self._name = new_name
+
+    def to(self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None):
+        """
+        Moves static Params for this component and its childs to the specified device and casts them to the specified data type.
+        """
+        # TODO I think we should make users specify a separate method so we dont override this behavior
+        for name in self._params.keys():
             param = self._params[name]
             if isinstance(param, torch.Tensor):
```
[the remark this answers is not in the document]
AlexandreAdam: Should be part of a future PR linked to [issue 48](https://github.com/Ciela-Institute/caustic/issues/48)

caustic/parametrized.py
```diff
@@ -24,124 +25,84 @@ class Parametrized:
...
-        # Propagate up through all parents
-        for parent in self._parents:
-            if not parent._can_add(p):
-                return False
-
-        return True
-
-    def _merge_descendants_up(self, p: "Parametrized"):
+    @staticmethod
+    def _generate_unique_name(name, module_names):
```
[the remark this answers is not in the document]
AlexandreAdam: Yes it might be an issue. Right now I assume that any module with the same name which do not have a direct edge to the current module are actually the same module. We will have such a situation when, for example, we do
```
class Sim(Simulator):
    def __init__(self):
        self.cosmo = FlatLambdaCDM()
        self.len = EPL(self.cosmo)
```

The DAG structure is
```

Sim -> FlatLambdaCDM
        -> EPL -> FlatLambdaCDM
```
We want to make sure FlatLambdaCDM is not renamed. Looking beyond the childs of Sim is a bit problematic and will cause FlatLambdaCDM to be tagged as a name collision when it is just a duplicate of the same noe. There might be another solution and mine certainly is not perfect. The code in place at least supports the most basic DAG structure (like the one above) with the correct behavior.

caustic/parametrized.py
AlexandreAdam: Now that I think about, `if module._childs` is even more general

caustic/parametrized.py
AlexandreAdam: Yes! I used the attributes, which I prefer to dict data access :).

test/test_namespace_dict.py
AlexandreAdam: Isn't that what the test with "Keep me" is doing?

ConnorStoneAstro: you're right, I just missed that

test/test_parametrized.py
AlexandreAdam: Yes I can, and what I have is essentially that. But x needed to be sent to tensor first so that I could test what happens. 

The way you constructed your example, I would have to send all lists (cosmo_params, lens_params and source_params) to tensor separately, which I simply did in a single line (line 65).

test/test_parametrized.py
AlexandreAdam: I think I forgot to remove the previous line, or wanted to test something. I'll remove the extra line.

test/test_parametrized.py
AlexandreAdam: I added a check for the shape. Checking the type should be part of the PR resolving [issue 48](https://github.com/Ciela-Institute/caustic/issues/48)

caustic/namespace_dict.py
```diff
@@ -0,0 +1,178 @@
...
+    def collapse(self) -> NamespaceDict:
+        """
+        Flatten the nested dictionary and collapse keys into the first level 
+        of the NamespaceDict
+        
+        Returns:
+            NamespaceDict: Flattened dictionary as a NamespaceDict
+        """
+        flattened_dict = NamespaceDict()
+        def _flatten_dict(dictionary, parent_key=""):
```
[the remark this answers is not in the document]
AlexandreAdam: Yes that is the point, see comment below. flatten should be used to avoid name collision.

ConnorStoneAstro (approved):

Issue #54, opened 2023-07-09
PseudoJaffe double naming of convergence_0


AlexandreAdam:
Why is `convergence_0` a parameter of the `PseudoJaffe` module and also a static method of the module? We should rename one or the other.

Issue #53, opened 2023-07-08
Cannot vmap through Simulator with Pixelated source


AlexandreAdam:
# Description of the issue
The problem is due to how `caustic.utils.interp2d` is written. When trying to vmap a simulator with a pixelated source, I get the following torch internal error 
```
RuntimeError: vmap: We do not support batching operators that can support dynamic shape. Attempting to batch over indexing with a boolean mask.
```
In our code, this error occurs specifically at line 279 of `caustic.utils`
```
if padding_mode == "zeros":  # else padding_mode == "extrapolate"
    result[idxs_out_of_bounds] = torch.zeros_like(result[idxs_out_of_bounds])
```

This can easily be replicated with the following code
```
from caustic import Simulator, EPL, FlatLambdaCDM, Pixelated
from caustic.utils import get_meshgrid
from torch import vmap # torch 2.0

n_pix = 20
class Sim(Simulator):
    def __init__(self, name="test"):
        super().__init__(name)
        pixel_scale = 0.04
        fov = n_pix * pixel_scale
        z_l = 0.5
        self.z_s = torch.tensor(1.0)
        self.cosmo = FlatLambdaCDM(h0=0.7 if cosmo_static else None, name="cosmo")
        self.epl = EPL(self.cosmo, z_l=z_l, name="lens")
        self.source = Pixelated(x0=0., y0=0., pixelscale=pixel_scale/2, image_shape=[n_pix, n_pix])
        self.thx, self.thy = get_meshgrid(pixel_scale, n_pix, n_pix)
        self.n_pix = n_pix

    def forward(self, params):
        alphax, alphay = self.epl.reduced_deflection_angle(x=self.thx, y=self.thy, z_s=self.z_s, params=params) 
        bx = self.thx - alphax
        by = self.thy - alphay
        return self.source.brightness(bx, by, params)

# default cosmo params
h0 = torch.tensor([0.68, 0.75])

# default lens params 
x0 = torch.tensor([0, 0.1])
y0 = torch.tensor([0, 0.1])
q = torch.tensor([0.9, 0.8])
phi = torch.tensor([-0.56, 0.8])
b = torch.tensor([1.5, 1.2])
t = torch.tensor([1.2, 1.0])

source = [torch.randn([2, n_pix, n_pix])]

cosmo_params = [h0]
lens_params = [x0, y0, q, phi, b, t]
x = [cosmo_params, lens_params, source]
sim = Sim()
vmap(sim)(x)
```

[...]

PR #56, opened 2023-07-11
Batch, dtypes and to method

AlexandreAdam:
# Features
- Fix for batching (`vmap`) a simulator with a Pixelated source. The fix is in interp2d, line 279. This fix [Issue 53](https://github.com/Ciela-Institute/caustic/issues/53).
- Added the possibility to set a value (None or a Tensor) to a parameter at will (added a setter method to Parameter and refactored the setattr of Parametrized)
- Added test for the `to` method and verified that it works as intended. This test validates that the recursive algorithm works and that module do not necessarily have to implement a to method. This is a fix for [Issue 48](https://github.com/Ciela-Institute/caustic/issues/48)
- Made sure that default `dtype` in Paramater is `torch.float32`
- Also proposed a fix for [Issue 54](https://github.com/Ciela-Institute/caustic/issues/54) for the PseudoJaffe module.
- Added a `dtype` property to Parameter.

### DIFF
--- a/caustic/parameter.py
+++ b/caustic/parameter.py
@@ -1,7 +1,9 @@
 from typing import Optional, Union
+from collections import OrderedDict
 
 import torch
 from torch import Tensor
+from caustic.namespace_dict import NamespaceDict
 
 __all__ = ("Parameter",)
 
@@ -31,6 +33,8 @@ class Parameter:
             ValueError: If both value and shape are None, or if shape is provided and doesn't match the shape of the value.
         """
         # Must assign one of value or shape
+        self._parents = NamespaceDict()
+        self._childs = None # A Parameter will always be a leaf node of the DAG
         self._value = value
         if value is None:
             if shape is None:
--- a/caustic/parameter.py
+++ b/caustic/parameter.py
@@ -22,8 +22,6 @@ class Parameter:
         self, value: Optional[Union[Tensor, float]] = None, shape: Optional[tuple[int, ...]] = ()
     ):
         # Must assign one of value or shape
-        self._parents = NamespaceDict()
-        self._childs = None # A Parameter will always be a leaf node of the DAG
         if value is None:
             if shape is None:
                 raise ValueError("If value is None, a shape must be provided")
@@ -72,9 +70,7 @@ class Parameter:
     def set_static(self):
         self.value = None
 
-    def to(
-        self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None
-    ):
+    def to(self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None):
         """
         Moves and/or casts the values of the parameter.
 
@@ -84,6 +80,7 @@ class Parameter:
         """
         if self.static:
             self.value = self._value.to(device=device, dtype=dtype)
+        return self
 
     def __repr__(self) -> str:
         if self.static:
--- a/caustic/parameter.py
+++ b/caustic/parameter.py
@@ -1,5 +1,4 @@
 from typing import Optional, Union
-from collections import OrderedDict
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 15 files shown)


caustic/parameter.py
```diff
@@ -20,92 +21,69 @@ class Parameter:
...
-        self._value = value
         if value is None:
             if shape is None:
                 raise ValueError("If value is None, a shape must be provided")
+            if not isinstance(shape, tuple):
+                raise TypeError("The shape of a parameter must be a tuple")
             self._shape = shape
         else:
-            value = torch.as_tensor(value)
+            value = torch.as_tensor(value).float() # set float32 as default dtype here
```
ConnorStoneAstro, line 32: no need to duplicate work, just do self.value = value and let the value setter handle the typing

caustic/parameter.py
```diff
@@ -20,92 +21,69 @@ class Parameter:
...
-
-        Returns:
-            Optional[Tensor]: The value of the parameter, or None if the parameter is dynamic.
-        """
         return self._value
+    
+    @value.setter
+    def value(self, value: Union[None, Tensor, float]):
+        if value is not None:
+            value = torch.as_tensor(value)
```
ConnorStoneAstro, line 56: are we going to enforce float32? if so this should be `torch.as_tensor(value, dtype = torch.float32)`

caustic/simulator.py
```diff
@@ -17,4 +17,11 @@ class Simulator(Parametrized):
     """
 
     def __call__(self, *args, **kwargs):
-        return self.forward(self.pack(args[0]), *args[1:], **kwargs)
+        packed_args = self.pack(args[0])
+        return self.forward(packed_args, *args[1:], **kwargs)
+        # packed_args, batched = self.pack(args[0])
+        # out = self.forward(packed_args, *args[1:], **kwargs)
```
ConnorStoneAstro, line 23: why is all of this commented out? Is it no longer needed?

test/test_parametrized.py
```diff
@@ -1,6 +1,10 @@
 import torch
+from torch import vmap
+import numpy as np
 from caustic import Simulator, EPL, Sersic, FlatLambdaCDM
-from caustic.utils import get_meshgrid
+from caustic.parameter import Parameter
+from utils import setup_image_simulator, setup_simulator
```
ConnorStoneAstro, line 6: `setup_image_simulator` is not used

test/test_pixelated.py
```diff
@@ -0,0 +1,12 @@
+# from caustic import Pixelated
+# from caustic.utils import get_meshgrid
+
+
+# TODO
+# def test_2d_pixelated_brightness():
```
ConnorStoneAstro, line 6: does this need to be done now?

caustic/parameter.py
AlexandreAdam: I don't think we should enforce it in the value setter. After all, we might have a situation where float16 is set for the whole simulator and user pass a float16 value. Then enforcing float32 in the value setter will break the code in an unintuitive way.

caustic/parameter.py
AlexandreAdam: The init code is slightly different from the value setter, both because of this default, and also creating the private attributes _shape for the first time (setter never redefines the shape, which is important).

caustic/simulator.py
AlexandreAdam: Ah, forgot this here when testing ideas with batch dimension.

caustic/parameter.py
ConnorStoneAstro: In that case it seems strange to set the dtype in the init as well. maybe leave it as `value = torch.as_tensor(value)`?

AlexandreAdam: Agreed, I removed this default.

ConnorStoneAstro (approved):

PR #57, opened 2023-07-11
refactoring for unpack decorator

ConnorStoneAstro:
Unpacking decorator creates uniform interface, more flexibility for new users, and removes unnecessary lines from method internals.

### DIFF
--- a/caustic/lenses/epl.py
+++ b/caustic/lenses/epl.py
@@ -4,7 +4,7 @@ import torch
 from torch import Tensor
 
 from ..cosmology import Cosmology
-from ..utils import derotate, translate_rotate, north
+from ..utils import derotate, translate_rotate
 from .base import ThinLens
 from ..parametrized import unpack
 
@@ -90,7 +90,7 @@ class EPL(ThinLens):
         Returns:
             tuple[Tensor, Tensor]: Reduced deflection angles in the x and y directions.
         """
-        x, y = translate_rotate(x, y, x0, y0, phi - north)
+        x, y = translate_rotate(x, y, x0, y0, phi)
 
         # follow Tessore et al 2015 (eq. 5)
         z = q * x + y * 1j
@@ -102,7 +102,7 @@ class EPL(ThinLens):
 
         alpha_real = torch.nan_to_num(alpha_c.real, posinf=10**10, neginf=-(10**10))
         alpha_imag = torch.nan_to_num(alpha_c.imag, posinf=10**10, neginf=-(10**10))
-        return derotate(alpha_real, alpha_imag, phi - north)
+        return derotate(alpha_real, alpha_imag, phi)
 
     def _r_omega(self, z, t, q):
         """
@@ -148,8 +148,8 @@ class EPL(ThinLens):
             Tensor: The lensing potential.
         """
         ax, ay = self.reduced_deflection_angle(x, y, z_s, params)
-        ax, ay = derotate(ax, ay, -(phi - north))
-        x, y = translate_rotate(x, y, x0, y0, phi - north)
+        ax, ay = derotate(ax, ay, -phi)
+        x, y = translate_rotate(x, y, x0, y0, phi)
         return (x * ax + y * ay) / (2 - t)
 
     @unpack(3)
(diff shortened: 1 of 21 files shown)


ConnorStoneAstro:
I have introduced a slight degeneracy. In the situation where there are N leading args, say 2, and one parameter you can get a scenario like this: `obj.method(arg1, arg2, None)` and the `None` will be interpreted as an empty packed object instead of trying to set the parameter equal to `None`. This should never actually be a problem since any time you want to set a parameter it should be a tensor.

ConnorStoneAstro:
Also, the decorator now catches args and kwargs, so feel free to use either, so long as the args are in the right order everything else can mix and match.

caustic/sources/sersic.py
```diff
@@ -92,9 +94,7 @@ def brightness(self, x, y, params: Optional["Packed"] = None):
             If `lenstronomy_k_mode` is True, we use the approximation from Lenstronomy, 
             otherwise, we use the approximation from Ciotti & Bertin (1999).
         """
-        x0, y0, q, phi, n, Re, Ie = self.unpack(params)
-
-        x, y = translate_rotate(x, y, x0, y0, phi)
+        x, y = translate_rotate(x, y, x0, y0, phi - north)
```
AlexandreAdam, line 97: Is this the best approach? I think we should do this internally , in the function `translate_rotate`. Otherwise we have to remember to subtract the north everytime we use tranlsate_rotate in future modules.

caustic/lenses/epl.py
```diff
@@ -102,7 +102,7 @@ def reduced_deflection_angle(
 
         alpha_real = torch.nan_to_num(alpha_c.real, posinf=10**10, neginf=-(10**10))
         alpha_imag = torch.nan_to_num(alpha_c.imag, posinf=10**10, neginf=-(10**10))
-        return derotate(alpha_real, alpha_imag, phi)
+        return derotate(alpha_real, alpha_imag, phi - north)
```
AlexandreAdam, line 105: Same comment as the one for translate_rotate.

caustic/lenses/epl.py
```diff
@@ -88,9 +90,7 @@ def reduced_deflection_angle(
         Returns:
             tuple[Tensor, Tensor]: Reduced deflection angles in the x and y directions.
         """
-        z_l, x0, y0, q, phi, b, t = self.unpack(params)
-
-        x, y = translate_rotate(x, y, x0, y0, phi)
+        x, y = translate_rotate(x, y, x0, y0, phi - north)
```
AlexandreAdam, line 93: North should be subtracted inside `translate_rotate`, not in the call.

caustic/lenses/epl.py
```diff
@@ -168,8 +168,6 @@ def convergence(
         Returns:
             Tensor: The convergence of the lens.
         """
-        z_l, x0, y0, q, phi, b, t = self.unpack(params)
-
-        x, y = translate_rotate(x, y, x0, y0, phi)
+        x, y = translate_rotate(x, y, x0, y0, phi - north)
```
AlexandreAdam, line 171: Same comment.

AlexandreAdam (changes requested):
Awsome PR, only two minor change requested.

caustic/sources/sersic.py
ConnorStoneAstro: I think this is the more general way to do it. That way calling `translate_rotate(x, y, x0, x0)` will just do translation with no rotation. If we have the `- NORTH` in the function then the default behaviour is to rotate by 90 deg which may be confusing. Also, are there any situations where we want a rotation, but don't need/care about orientation relative to north?

ConnorStoneAstro: Never mind, I can make it work and then yes it will be a lot cleaner.

AlexandreAdam:
> Also, the decorator now catches args and kwargs, so feel free to use either, so long as the args are in the right order everything else can mix and match.

This actually is awsome for using `vmap` on the simulator in more sophisticated ways.

AlexandreAdam (approved):
Looks good!

PR #59, opened 2023-07-14
unpack now in base as well. All functions catch kwargs too

ConnorStoneAstro:


AlexandreAdam (approved):

AlexandreAdam:
That looks good to me, I thought kwargs were in the previous PR but now it's a bit more clear. I think in future PR we should add some tests around this to constrain our target behavior better.

PR #63, opened 2023-08-02
Pixel scale quick fix

AlexandreAdam:
Changed the pixelated classes to be defined by a pixel scale parameters (in arcseconds) instead of field of view.


ConnorStoneAstro (approved):
Looks good!

PR #64, opened 2023-08-03
Fixed graph viewer parameter bug.

AlexandreAdam:
Please update the pypi version as we make merge requests, we will need this for Astromatic.


ConnorStoneAstro (approved):
Good catch!

ConnorStoneAstro:
> Please update the pypi version as we make merge requests, we will need this for Astromatic.

Ok, I'll wait until Friday afternoon to update pypi in case there are any more updates.

PR #65, opened 2023-08-04
base lens can now compute deflection angle jacobian

ConnorStoneAstro:
The Jacobian of the deflection angle can be used to construct critical curves and caustics. Since `caustic` is autodifferentiable, this is really easy!


AlexandreAdam (approved):
This works well!

Issue #6, opened 2022-12-09
Implement lens equation solver


adam-coogan:
This is needed for determining the positions of images of point-like sources (e.g. quasars). I'm imagining something like this:
```python
from typing import Union
from caustic.lenses.base import AbstractThinLens, AbstractThickLens

def forward_raytrace(
    thx0_src,
    thy0_src,
    lens: Union[AbstractThinLens, AbstractThickLens],
    z_l,
    z_s,
    cosmology,
    *args_lens,
    **kwargs_lens,
):
    def fn(thx, thy):
        """
        Function to minimize: distance between reverse ray-traced points and source's
        position.
        """
        bx, by = lens.raytrace(thx, thy, z_l, z_s, cosmology, *args_lens, **kwargs_lens)
        return (bx - thx0_src)**2 + (by - thy0_src)**2
    
    # TODO: implement. Could get guesses for image positions and fit from there.
    ...
```
It'd be good to write this as a standalone function first and then think about turning it into a method for the base lens classes.

For reference, Lenstronomy handles this in the [LensEquationSolver](https://lenstronomy.readthedocs.io/en/latest/lenstronomy.LensModel.Solver.html) class, which contains a generic solver and a faster one that works with EPL + shear models only. Ideally we'd implement both approaches.

adam-coogan:
This is roughly working! The main issue is that [levmarq-torch](https://github.com/adam-coogan/levmarq-torch) is not batched very well. It also needs to be added as a dependency.

ConnorStoneAstro:
Closed by #67

PR #67, opened 2023-08-11
Forward raytrace in base lens class

ConnorStoneAstro:
Forward raytrace is now a class method for lenses. It's format is similar to raytrace, though it goes the other way. Also only a single point can be forward raytraced at a time due to the variable output shape.


ConnorStoneAstro:
Closes #6

PR #68, opened 2023-08-11
Change deflection angle jacobian to lens equation jacobian

ConnorStoneAstro:


test/test_sie.py
```diff
@@ -45,7 +45,7 @@ def test_jacobian():
     # Parameters
     z_s = torch.tensor(1.2)
     x = torch.tensor([0.5, 0.912, -0.442, 0.7, pi / 3, 1.4])
-    J = lens.jacobian_reduced_deflection_angle(thx, thy, z_s, lens.pack(x))
+    J = lens.jacobian_lens_equation(thx, thy, z_s, lens.pack(x))
```
AlexandreAdam, line 48: Do we have a test comparing the finite difference and the autograd method? At least adding a check that we can call the method and that it works woud be appropriate.

AlexandreAdam (changes requested):
This looks very good! After the unit test this pull request will be ready to merge.

test/test_sie.py
ConnorStoneAstro: Good idea! Adding that now

AlexandreAdam (approved):
Great work!

PR #71, opened 2023-08-22
Cosmology now a full module with optional Phytorch

mjyb16:
If I did not break anything regarding Caustic design principles, I think everything is ready to go. I tested creating a lens with the new Phytorch differentiable cosmology and it works, and the distances (especially gradients) are more accurate with Phytorch.


mjyb16:
Looks like I need to modify the tests...I will try again to see if that can be fixed.

PR #76, opened 2023-09-06
Adding a docs page to caustic

ConnorStoneAstro:
It's not much, but its an honest start


ConnorStoneAstro:
I've matched up with my other repos as much as I can for the github workflow, but I may have to make adjustments when it actually gets merged to main

ConnorStoneAstro:
To help everything compile correctly Ive updated from .toml to a setup.py file. I think this is the more standard way to build a python package, but I'm not sure (I'm certainly more familiar with it)

andreasfilipp (approved):
I guess that's a good start! Thanks Connor

PR #77, opened 2023-09-07
Update tnfw.py

andreasfilipp:
update projected_mass to mass_enclosed_2d to match pseudoJaffe profile. Also the input radius to calculate enclosed mass is now in arcsec (to be consistent with pseudoJaffe profile).


ConnorStoneAstro (approved):
Looks good!

PR #84, opened 2023-10-08
Bug fixes 20231008

AlexandreAdam:
- A minor bug fix in the `Parameter.__repr__` method.
- An important bug fix in `Cosmology.__init__` not sending private tensors `device`.
- An important bug fixed in the `@unpack` decorator not handling the unpacking correctly list or dict `params` passed as an argument or a keyword argument. This issue was found using the `Multiplane` lens, and an appropriate test was created using this object. This issue was missed in previous tests since most of them use the `Simulator` class. The `Simulator.__call__` method does not use the unpack decorator, rather it calls `Parametrized.pack` directly.
- Loosened the requirements file

### DIFF
--- a/test/test_cosmology.py
+++ b/test/test_cosmology.py
@@ -35,6 +35,9 @@ def test_comoving_dist():
 
 def test_to_method_flatlambdacdm():
     cosmo = CausticFlatLambdaCDM()
+    # Make sure private tensors are created on float32 by default
+    assert cosmo._comoving_distance_helper_x_grid.dtype == torch.float32
+    assert cosmo._comoving_distance_helper_y_grid.dtype == torch.float32
     cosmo.to(dtype=torch.float64)
     # Make sure distance helper get sent to proper dtype and device
     assert cosmo._comoving_distance_helper_x_grid.dtype == torch.float64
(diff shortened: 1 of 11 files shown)


test/test_cosmology.py
```diff
@@ -33,5 +33,13 @@ def test_comoving_dist():
         assert np.allclose(vals, vals_ref, rtol, atol)
 
 
+def test_to_method_flatlambdacdm():
+    cosmo = CausticFlatLambdaCDM()
```
ConnorStoneAstro, line 37: I think everything defaults to float32, but perhaps you should explicitly construct it as float32, test that everything is float32, then call `to` and test that everything is float64

AlexandreAdam: This test was meant only to make sure private tensor of FlatLambdaCDM are sent to dtype and device. Please refer to `test_to_method` in the file `test_parametrized` for actual `.to()` method test. 

For this specific case, the private tensors are explictly constructed as `float32` tensors in the `__init__` method. See `FlatLambdaCDM` class.

ConnorStoneAstro (approved):

ConnorStoneAstro (approved):

PR #89, opened 2023-10-20
move source to light

### LINKED ISSUE #87: Nomenclature change `sources` -> `light`
I think we should change the `sources` nomenclature to `light` as in a "sersic light model" just like how one discusses as "SIE lens model". The light models can be used as sources, but also as the lens light, or even just an interloper. In principle the "source" models can represent anything which generates light.

We should also add a new directory for PSF models since we currently have no internal treatment for PSF.

ConnorStoneAstro:
This change simply makes light models more generally applicable (the name not the functionality). Since they are used for the source, but also for the lens light and interlopers in principle.

This closes #87


AlexandreAdam:
That's a good idea. I would merge this into main but wait for other pull requests before bumping the version on pypi to include more change. This way we would package a bunch of changes together. 

Thus I think we should push this change into dev before sending it to main.

AlexandreAdam (approved):

ConnorStoneAstro:
Oh, I see. That might be a good workflow in general to merge into dev then periodically merge dev to main and update the version.

caustic/lenses/base.py
```diff
@@ -14,7 +14,13 @@
 
 __all__ = ("ThinLens", "ThickLens")
 
-class ThickLens(Parametrized):
+class Lens(Parametrized):
```
AlexandreAdam, line 17: Why add this class? It does nothing in the inheritance tree. I think we should add that only when we need it, for example if we have control flow common to ThickLens and ThinLenses

AlexandreAdam (changes requested):
This looks good overall.

caustic/lenses/base.py
ConnorStoneAstro: There is actually some stuff I was planning to move into this class since it is shared by both ThinLens and ThickLens. Also, it is nice in general to be able to test if an object is a Lens object as opposed to say a Light object, Cosmology, etc.

But I can move this to a future PR if that's easier.

AlexandreAdam (approved):

PR #94, opened 2023-10-24
Add interactive demo for lens source

ConnorStoneAstro:
This is a demo for the purposes of teaching gravitational lensing. It runs by calling: 
```
import caustic
caustic.demo.lens_source()
```

and will then show an interactive figure. I tried to get it working in a jupyter notebook but it was too buggy/choppy. Open to ideas.


ConnorStoneAstro:
@mjyb16 is working on a superior demo system, so this PR is no longer needed

PR #112, opened 2023-12-08
test: Adding test for quick test

ConnorStoneAstro:
This just makes sure that the quick test runs, and makes sure that our code coverage doesnt get angry

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -3,6 +3,7 @@ import numpy as np
 
 from caustics.cosmology import FlatLambdaCDM
 from caustics.lenses import SIE
+from caustics import test as mini_test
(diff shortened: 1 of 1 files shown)


tests/test_base.py
```diff
@@ -34,3 +35,6 @@ def test():
 
     assert torch.all((sp_x - bx).abs() < 1e-3)
     assert torch.all((sp_y - by).abs() < 1e-3)
+
+def test_quicktest():
```
lsetiawan, line 39:
```suggestion
def test_quicktest():
    """Testing the test for install validation"""
```

lsetiawan (approved):

PR #124, opened 2023-12-11
style: Perform full linting and fixes to the whole codebase

lsetiawan:
Lint and fix the entire codebase to follow the PEP8 Guidelines. Currently the only one check that is turned off is the mypy, which fixes will happen in a separate PR. Pre-commit CI check should now pass.


.pre-commit-config.yaml
```diff
@@ -1,28 +1,26 @@
 exclude: |
   (?x)^(
-    tests/utils.py |
-    .devcontainer/
+    tests/utils.py
```
lsetiawan, line 3: This is the only file excluded because otherwise it will fail, since it doesn't follow the correct filename for test file, which should be like `test*.py`. Maybe testing utilities can move to the main code `tests` module?

ConnorStoneAstro: These utils are solely relevant for testing so to me it makes sense to keep them with the tests and just have a config line excluding them. What do you think @AlexandreAdam ?

src/caustics/__init__.py
```diff
@@ -1,16 +1,23 @@
...
-from .tests import *
+from . import constants, lenses, cosmology, packed, parametrized, light, utils, sims
+from .tests import test
 
 # from .demo import *
 
 __version__ = VERSION
 __author__ = "Ciela"
+
+__all__ = [
```
ConnorStoneAstro, line 11: Is the `__all__` needed when it references everything? Also, will `__version__` and `__author__` still be accessible if they are not in `__all__`?

.pre-commit-config.yaml
lsetiawan: Great! It that's the case, then sounds good to me 😄

ConnorStoneAstro (commented):
I see a lot of `from x import *` are removed, which is fine in general. However it means that maintenance is more work since any new function must be traced through the import hierarchy and added to all the imports and `__all__`'s. What is the benefit of doing things this way?

lsetiawan:
Great question @ConnorStoneAstro!

PEP8 Standard states the following:

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn’t known in advance).

This is typically not a recommended way to import functions and classes as it is very implicit and hard to understand what is actually being imported. One great article that really hammer on this ambiguity can be found at: https://internetworking.dev/why-it-is-bad-idea-to-use-wildcard-based-imports/

> However it means that maintenance is more work since any new function must be traced through the import hierarchy and added to all the imports and __all__'s. What is the benefit of doing things this way?

To this point, really what you need is to import any new *public* functions in a module to the module's `__init__.py`. I notice that per class for example: `src/caustics/light/base.py` you have an `__all__`, this is not necessary and better to put it in the module level. Think of this module level `__all__` as specifying exactly what functions or classes that you'd like to expose to your users. I think the benefit here is really readability and easier to follow the code and knowing what's getting exposed vs what's hidden. This allows both users and outside developers to easily understand the important parts of your code.

Normally when programming Python it is always better to be explicit rather than implicit to avoid any confusions, hence their PEP8 above.

src/caustics/__init__.py
lsetiawan: This exposes the modules that is public. For example, as you may have notice that originally, `data` module is not imported, so users wouldn't be able to just do `import caustics; caustics.data.HDF5Dataset`. If `data` was imported here and added to `__all__` then you would be able to do that.

At the end of the day, really this part of the thought process of what your final API is like. If it's expected that users can access modules straight from `caustics` imports, then this is what needs to happen. If a user is expected to import the functions from each individual module, then you don't need to specify it, so in the case of `data` since it's not in `__all__` users would have to do like `from caustics.data import HDF5Dataset`.

ConnorStoneAstro (approved):
This looks great!

PR #126, opened 2023-12-12
docs: extend timeout to allow matplotlib to build font library

ConnorStoneAstro:
Matplotlib is timing out our automated jupyter-notebook builds while trying to make a font library. The new timeout limit is set to 2 minutes to give it a chance to build

### DIFF
--- a/.codespell-whitelist
+++ b/.codespell-whitelist
@@ -1,0 +1,4 @@
+sie
+childs
+dout
+din
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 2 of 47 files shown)


ConnorStoneAstro:
Should we push this directly to main? since it is crashing there too

lsetiawan:
> Should we push this directly to main? since it is crashing there too

Let's do that, this is a great patch.

lsetiawan (approved):

lsetiawan (changes requested):
Welp, since this is branched of dev looks like you'll have to create a new PR since this has logs from current `dev` rather than from `main`.

lsetiawan:
I think this other PR will fix the issue at hand: https://github.com/Ciela-Institute/caustics/pull/128

PR #127, opened 2023-12-12
docs: Update tutorials to call modules with new paths

ConnorStoneAstro:
With the new import hierarchy, some links in the tutorials were broken. Now they link correctly.


lsetiawan (approved):
This is really great. It's much more clear as to where each function and classes are coming from. Though it can get quite long, so we'd really have to think about how to improve this in the new API design work.

PR #128, opened 2023-12-12
ci: Add matplotlib import in RTD pre_build

lsetiawan:
Add matplotlib import in the beginning to avoid creation of font cache during building of jupyter notebooks.


ConnorStoneAstro (approved):
That makes more sense than my fix! LGTM

PR #134, opened 2023-12-18
docs: Adding basic lensing example

ConnorStoneAstro:
A basic lensing example is added to the README. I also use this minimal example to build a new `BasicIntroduction` tutorial which I think is easier to follow.


README.md
```diff
@@ -25,6 +25,30 @@ Simply install caustics from PyPI:
...
+sie = caustics.lenses.SIE(
+    cosmology=C, z_l=0.5, x0=0.0, y0=0.0, q=0.4, phi=1.5708, b=1.0
+)
+src = caustics.light.Sersic(x0=-0.2, y0=0.0, q=0.6, phi=-0.785, n=1.5, Re=3.0, Ie=1.0)
+
+sim = caustics.sims.Lens_Source(
+    lens=sie, source=src, pixelscale=0.05, pixels_x=100, z_s=1.5
+)
+
+plt.imshow(sim(quad_level=3).detach().cpu().numpy(), origin="lower")
```
AlexandreAdam, line 45: I don't know that we should put quad_level in the basic example. I think it would be more clear without.

docs/source/intro.md
```diff
@@ -20,6 +20,30 @@ are ready to go! You can check out the tutorials afterwards to see some of
...
+sie = caustics.lenses.SIE(
+    cosmology=C, z_l=0.5, x0=0.0, y0=0.0, q=0.4, phi=1.5708, b=1.0
+)
+src = caustics.light.Sersic(x0=-0.2, y0=0.0, q=0.6, phi=-0.785, n=1.5, Re=3.0, Ie=1.0)
+
+sim = caustics.sims.Lens_Source(
+    lens=sie, source=src, pixelscale=0.05, pixels_x=100, z_s=1.5
+)
+
+plt.imshow(sim(quad_level=3).detach().cpu().numpy(), origin="lower")
```
AlexandreAdam, line 40: Same as above

README.md
```diff
@@ -25,6 +25,30 @@ Simply install caustics from PyPI:
...
 ```
 
+## Minimal Example
+
+```python
+import matplotlib.pyplot as plt
+import caustics
+
+C = caustics.cosmology.FlatLambdaCDM()
+
```
AlexandreAdam, line 35: I perhaps take issue with this basic example. It's the wrong way to use caustic even if in terms of code it is perhaps the simplest. 

A user copying and implementing this code in their pipeline would find caustics slow and clunky

docs/source/intro.md
```diff
@@ -20,6 +20,30 @@ are ready to go! You can check out the tutorials afterwards to see some of
...
+plt.axis("off")
+plt.show()
+```
+
+![Caustics lensed image](../../media/minimal_example.png)
+
 ### Contents
```
AlexandreAdam, line 47: Add vmap example

docs/source/intro.md
```diff
@@ -20,6 +20,30 @@ are ready to go! You can check out the tutorials afterwards to see some of
...
```
AlexandreAdam, line 48: Add batch example

AlexandreAdam (changes requested):

README.md
ConnorStoneAstro: I would agree, except if we are assuming people will copy/paste this example to work with, we would want them to get accurate fluxes. The difference between direct sampling and `quad_level=3` is actually quite significant... Someone copy/pasting may not think about their integration depth.

README.md
ConnorStoneAstro: I agree, I'll change it to be more in line with proper style

PR #136, opened 2023-12-18
refactor: new parametrized unpack decorator requires function signature with kwargs

ConnorStoneAstro:
The `unpack` decorator now requires function signatures to be of the form: 
```python
func(self, arg1, arg2, ..., *args, params = None, param1 = None, param2 = None, ..., kwarg1 = val1, kwarg2 = val2, **kwargs)
```
This allows for some notable benefits:

- `@unapck` is all that's needed as the decorator, no longer `@unpack(<number>)`
- Can now identify which parameter is missing and provide helpful error message of the form: `KeyError: "Missing parameter 'x0' in method 'jacobian_lens_equation' of module 'sie'"`
- kwargs that aren't parameters can now get passed reliably without crashing
- `unpack` decorator is much simpler
- It is aware of the dynamic parameters needed and so behaves nicer when mixed static/dynamic are used

The only case we lose is parameters passed as args like this:

```python
lens = SIE()
lens.raytrace(x,y,z_s,z_l,x0,y0,q,phi,b)
```

It would instead need to be:

```python
lens = SIE()
lens.raytrace(x,y,z_s,z_l=z_l,x0=x0,y0=y0,q=q,phi=phi,b=b)
```

and it would work.

Cases where everything is packed would still work, even passed as an arg, so long as it's the last arg:

```python
lens = SIE()
x = [1,1,1,1,1,1]
lens.raytrace(x,y,z_s, lens.pack(x))
```

That is just like before.


AlexandreAdam (commented):
Comments are about passing kwargs forward in method so that called method also have access to them

AlexandreAdam:
I think this is a very good PR. As discussed, all that is needed now is to implement the change in every method, make sure kwargs are passed forward. 

For documentation purposes, it would be nice if you could highlight some examples where this change impacts the API. In particular, how a user who wants to call directly a method without using the packed argument would interact with this feature and how it plays with dynamic/static parameters. Would make a neat jupyter notebook.

tests/test_parametrized.py
```diff
@@ -90,6 +90,47 @@ def test_unpack_some_modules_static():
...
 
 
+def test_pass_params_as_kwargs():
+    C = FlatLambdaCDM(name="cosmo")
+    lens = Point(cosmology=C)
+    thx, thy = torch.tensor(0.5), torch.tensor(0.5)
+    P = lens.potential(
+        thx,
+        thy,
+        z_s=torch.tensor(1.0),
```
AlexandreAdam, line 100: Does the method work if z_s and other parameters are not tensors?

tests/test_parametrized.py
```diff
@@ -90,6 +90,47 @@ def test_unpack_some_modules_static():
...
+        P = lens.potential(
+            thx,
+            thy,
+            z_s=torch.tensor(1.0),
+            z_l=torch.tensor(0.5),
+            x0=torch.tensor(0.0),
+            y0=torch.tensor(0.0),
+            th_ein=torch.tensor(1.0),
+        )
+
```
AlexandreAdam, line 119: Need to test with batched example

AlexandreAdam (changes requested):
Need to expand the scope of the tests

tests/test_parametrized.py
ConnorStoneAstro: I checked, they have to be tensors. It's only when assigning them during the class `__init__` that they can be floats that will be cast automatically. This is probably for performance reasons. Don't want to be casting to tensor on every call.

tests/test_parametrized.py
ConnorStoneAstro: I think the `test_batching` test already covers this. You probably wouldn't (couldn't?) vmap with all these kwargs.

AlexandreAdam: Yeah that is what I wanted to know, if the kwargs setup could support vmap. I think vmap specifically does not work with kwargs but can work with args.

ConnorStoneAstro: Yeah `vmap` wont work directly with kwargs, but you can write a wrapper if you really want. 
`lambda a,b,c: lens.func(a=a, b=b, c=c)` will turn args to kwargs if its really important. But I think anyone who is vmapping would just use the packed params like this: `lambda x: lens.func(lens.pack(x))` (just added a test for this and it works). Or if they are using a simulator then it packs internally so you can just do `vmap(sim)(x)`, this already has tests.

docs/source/tutorials/Parameters.ipynb
```diff
@@ -0,0 +1,247 @@
...
+    "        \"y0\": torch.tensor(-0.1),\n",
+    "        \"q\": torch.tensor(0.6),\n",
+    "        \"n\": torch.tensor(3.0),\n",
+    "        \"Ie\": torch.tensor(1.0),\n",
+    "    },\n",
+    "}\n",
+    "\n",
+    "fig, axarr = plt.subplots(1, 2, figsize=(16, 8))\n",
+    "axarr[0].imshow(\n",
+    "    sim(sim.pack(x_tens), quad_level=3).detach().cpu().numpy(), origin=\"lower\"\n",
```
AlexandreAdam, line 170: Change this to `sim(x)` instead. It's better that the user uses our built-in forward method instead of the pack method. The pack method should almso be considered as a private method.

docs/source/tutorials/Parameters.ipynb
```diff
@@ -0,0 +1,247 @@
...
+    ")\n",
+    "axarr[0].set_title(\"Simulator from tensor\")\n",
+    "axarr[1].imshow(\n",
+    "    sim(sim.pack(x_dict), quad_level=3).detach().cpu().numpy(), origin=\"lower\"\n",
```
AlexandreAdam, line 174: Same here

src/caustics/parametrized.py
```diff
@@ -370,10 +369,11 @@ def params(self) -> NestedNamespaceDict:
         dynamic = NestedNamespaceDict()
 
         def _get_params(module):
-            if module.module_params.static:
-                static[module.name] = module.module_params.static
-            if module.module_params.dynamic:
-                dynamic[module.name] = module.module_params.dynamic
+            MP = module.module_params
```
AlexandreAdam, line 372: I don't agree with this artificial change. Capital letters should not be used as name variable except for global variables stated with the import statements. Please revert this code to its original form.

tests/test_masssheet.py
```diff
@@ -16,8 +16,8 @@ def test():
 
     thx, thy = get_meshgrid(0.01, 10, 10)
 
-    ax, ay = lens.reduced_deflection_angle(thx, thy, z_s, *x)
+    ax, ay = lens.reduced_deflection_angle(thx, thy, z_s, lens.pack(x))
 
-    lens.potential(thx, thy, z_s, *x)
+    lens.potential(thx, thy, z_s, lens.pack(x))
```
AlexandreAdam, line 21: Is the use of the pack method necessary here?

tests/test_masssheet.py
```diff
@@ -16,8 +16,8 @@ def test():
...
 
-    lens.convergence(thx, thy, z_s, *x)
+    lens.convergence(thx, thy, z_s, lens.pack(x))
```
AlexandreAdam, line 23: Same as above

tests/test_multiplane.py
```diff
@@ -145,7 +145,7 @@ def test_params():
     multiplane_lens = Multiplane(cosmology=cosmology, lenses=planes)
     z_s = torch.tensor(z_s)
     x, y = get_meshgrid(pixel_size, 32, 32)
-    params = [torch.randn(pixels, pixels) for i in range(10)]
+    params = multiplane_lens.pack([torch.randn(pixels, pixels) for i in range(10)])
```
AlexandreAdam, line 148: Same as above

tests/test_multiplane.py
```diff
@@ -164,7 +164,9 @@ def test_params():
     )
 
     # Test that we can pass a dictionary
-    params = {f"plane_{p}": torch.randn(pixels, pixels) for p in range(n_planes)}
+    params = multiplane_lens.pack(
+        {f"plane_{p}": torch.randn(pixels, pixels) for p in range(n_planes)}
+    )
```
AlexandreAdam, line 169: Same as above

tests/test_parametrized.py
AlexandreAdam: I am a bit confused as to why the pack is necessary in this setup. Can we avoid it and simply have x?

AlexandreAdam (changes requested):
My comments are centered around the usage of the pack method, which should be considered more as a private method than a public method in caustics.

src/caustics/parametrized.py
ConnorStoneAstro: Good catch, changed to lower case. The other modification just helps reduce the number of calls to recursive functions that re-build the same dictionary multiple times. At some point we may want to go through all of `parametrized` and see where we can optimize the calls

AlexandreAdam (approved):
This is great!

PR #137, opened 2023-12-19
chore: add pre-commit badge

ConnorStoneAstro:
I like the shiny badges


mjyb16 (approved):
I also like shiny badges

PR #138, opened 2023-12-21
tests: Add consistency test between lensing potential and deflection angle, convergence

ConnorStoneAstro:
New test checks for  `grad(potential) = deflection` and for `laplacian(potential) = convergence`. This test has identified a few loose ends. 

- `PixelatedConvergence` doesn't pass either test and I'm not sure why. 
- `PseudoJaffe` also failed both tests, I think this is because of a unit conversion in the potential. 
- `NFW` and `TNFW` couldn't do the convergence test because autograd and vmap can't both work at the same time
- `Point` can't do the convergence test because it's convergence is a delta function which can't be tested
- `TNFW` failed the deflection angle test because of a unit conversion most likely. The units in the reference come out to Mpc which is not arcsec^2 

Otherwise, the lenses pass this consistency check, which is pretty good!

To make my life easier, I added a new `_null_params` class variable for each of the lenses which specifies some basic parameter values that are just reasonable choices.


ConnorStoneAstro:
Testing TNFW potential, the relative values for deflection angle (`ax / phi_ax`) do not change with perturbations in any of the TNFW parameters (`x0, y0, mass, scale_radius, tau`). This suggests it's purely a units problem for the potential.

ConnorStoneAstro:
I managed to reverse-engineer the solution for `TNFW` and `PseudoJaffe`. It turns out the potentials were "physical potentials" and so their derivatives were giving physical deflection angles. To ensure they are in line with the potentials in the other lens models, I switched them to reduced potentials. They now pass the potential tests.

ConnorStoneAstro:
The `PixelatedConvergence` now seems to work for the deflection field. This is only because I set the tolerance very low. For the convergence test I now realize that it can never work because the calculations are done via bilinear interpolation which always has zero second derivative. The bilinear interpolation is also probably why the deflection field through the potential is very inaccurate. I have left a very low tolerance test on, this way it will flag any major errors such as a units mismatch.

PR #140, opened 2023-12-29
fix: Fix codespace configuration

lsetiawan:
## Overview

The current codespace doesn't really install the necessary dependency for the software as the configuration doesn't run any onbuild steps. Additionally, the original built image with pytorch is too big for Github codespaces to actually spin up. This new configuration just pulls the bare minimum from `base-notebook` image, install a few apt packages, and then the actual development installation and setup for caustics happens after the codespace spins up. This setup now includes pre-commit setup.


.devcontainer/cpu/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+{
+  "name": "Caustics - CPU",
+  "build": {
+    "context": "../..",
+    "dockerfile": "../Dockerfile",
+    "args": {
+      "USERNAME": "vscode",
+      "BUILDKIT_INLINE_CACHE": "0",
+      "CLANG_VERSION": ""
+    }
```
lsetiawan, line 12:
```suggestion
    "dockerfile": "../Dockerfile",
    "args": {
      "CLANG_VERSION": ""
    }
```
No need for these args... old configuration for a different image.

ConnorStoneAstro (approved):
It's a bit above my head, but from your description of the PR this looks good!

PR #148, opened 2024-01-23
style: Activated Mypy Checks and Code Update for Compliance (#58)

uwcdc:
* fix: mypy compliance

* style: pre-commit fixes

---------

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,7 +55,6 @@ repos:
       - id: ruff
         args: ["--fix", "--show-fixes"]
 
-  # 2023-12-11: Not use mypy for now.
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: "v1.8.0"
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,14 +56,14 @@ repos:
         args: ["--fix", "--show-fixes"]
 
   # 2023-12-11: Not use mypy for now.
-  # - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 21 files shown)


.pre-commit-config.yaml
```diff
@@ -56,14 +56,14 @@ repos:
         args: ["--fix", "--show-fixes"]
 
   # 2023-12-11: Not use mypy for now.
```
ConnorStoneAstro, line 58: This comment can be deleted, right?

.pre-commit-config.yaml
```diff
@@ -56,14 +56,14 @@ repos:
...
-  #       files: src|tests
-  #       args: []
-  #       additional_dependencies:
-  #         - pytest
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v1.8.0"
+    hooks:
+      - id: mypy
+        files: src
+        args: ["--ignore-missing-imports"]
```
ConnorStoneAstro, line 64: Which imports need to be ignored?

ConnorStoneAstro (changes requested):
Very nice! Everything is so much clearer!

.pre-commit-config.yaml
lsetiawan: Mypy for some reason get's confused with imports. Since this is already being checked by ruff, we can ignore for mypy.

.pre-commit-config.yaml
lsetiawan, line 58:
```suggestion
```

ConnorStoneAstro (approved):
Looks good TM!

PR #155, opened 2024-02-07
Fixed imports related to issue 154 and updated requirements for dev

### LINKED ISSUE #154: Caustics import
I would like to be able to imports modules from caustics in the following way

```python
from caustics import NFW, FlatLambdaCDM
```

(like it was before v0.7.0 I believe)

rather than the current way

```python
from caustics.lenses import NFW
from caustics.cosmology import FlatLambdaCDM
```

The first way is more convenient, especially if I already know what I want from the caustics. Also, the first approach is very reminiscent of how somebody use functions from torch or numpy. Often one can simply import numpy and get all the function he needs from the root of the package rather than having to import many submodule, which worsen the user experience and makes it much less intuitive.

AlexandreAdam:
This PR closes #154


pyproject.toml
```diff
@@ -43,7 +43,9 @@ Issues = "https://github.com/Ciela-Institute/caustics/issues"
 
 [project.optional-dependencies]
 dev = [
-    "lenstronomy==1.11.1"
+    "numba==0.58", # This is because of lenstronomy broken install thanks
```
ConnorStoneAstro, line 46: These changes can be reverted. The latest dev fixes the numba issue

src/caustics/__init__.py
```diff
@@ -1,23 +1,11 @@
 from ._version import version as VERSION  # noqa
 
-from . import constants, lenses, cosmology, packed, parametrized, light, utils, sims
-from .tests import test
-
-# from .demo import *
+from .cosmology import *
```
ConnorStoneAstro, line 3: To match with proper import formatting, every object must be imported explicitly. So this should become:
`from .cosmology import Cosmology, FlatLambdaCDM, h0_default, critical_density_0_default, Om0_default`
The other imports should also write out everything explicitly. It's a pain, but it only needs to be done once and then it's a nice format.

tests/test_imports.py
ConnorStoneAstro, line 1: Is this file needed?

ConnorStoneAstro (changes requested):
This will be a good ease of use upgrade. But we need to follow proper python conventions and have explicit imports.

ConnorStoneAstro (approved):

PR #158, opened 2024-02-13
chore(deps): Add testing dependencies for optional 'dev'

lsetiawan:
## Overview

This PR adds pytest and pre-commit for the 'dev' optional dependencies to make it easier for the user to have the full suite of dependencies to then perform testing as well during development.


ConnorStoneAstro (approved):
Nice!

PR #160, opened 2024-02-14
test: Add DEVICE global variable to minimal tests 

lsetiawan:
## Overview

This PR just adds a default torch device on the minimal testing functionality for caustics. This is a beginning to the full blown tests.

### DIFF
--- a/src/caustics/tests.py
+++ b/src/caustics/tests.py
@@ -10,6 +10,10 @@ from caustics.utils import gaussian, get_meshgrid
 
 __all__ = ["test"]
 
+TORCH_DEVICE = (
+    torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+)
+
 
 def _test_simulator_runs():
     # Model
@@ -32,7 +36,7 @@ def _test_simulator_runs():
         name="lenslight", x0=0.0, y0=0.01, q=0.7, phi=pi / 4, n=3.0, Re=0.7, Ie=1.0
     )
 
-    psf = gaussian(0.05, 11, 11, 0.2, upsample=2)
+    psf = gaussian(0.05, 11, 11, 0.2, upsample=2, device=TORCH_DEVICE)
 
     sim = Lens_Source(
         lens=lensmass,
@@ -43,6 +47,9 @@ def _test_simulator_runs():
         psf=psf,
         z_s=2.0,
     )
+
+    # Send to device
+    sim = sim.to(TORCH_DEVICE)
 
     assert torch.all(torch.isfinite(sim()))
     assert torch.all(
@@ -95,11 +102,14 @@ def _test_jacobian_autograd_vs_finitediff():
     # Models
     cosmology = FlatLambdaCDM(name="cosmo")
     lens = SIE(name="sie", cosmology=cosmology)
-    thx, thy = get_meshgrid(0.01, 20, 20)
+    thx, thy = get_meshgrid(0.01, 20, 20, device=TORCH_DEVICE)
 
     # Parameters
-    z_s = torch.tensor(1.2)
-    x = torch.tensor([0.5, 0.912, -0.442, 0.7, pi / 3, 1.4])
+    z_s = torch.tensor(1.2, device=TORCH_DEVICE)
+    x = torch.tensor([0.5, 0.912, -0.442, 0.7, pi / 3, 1.4], device=TORCH_DEVICE)
+
+    # Send to device
+    lens = lens.to(TORCH_DEVICE)
 
     # Evaluate Jacobian
     J_autograd = lens.jacobian_lens_equation(thx, thy, z_s, lens.pack(x))
@@ -115,9 +125,9 @@ def _test_jacobian_autograd_vs_finitediff():
 
 def _test_multiplane_jacobian():
     # Setup
-    z_s = torch.tensor(1.5, dtype=torch.float32)
+    z_s = torch.tensor(1.5, dtype=torch.float32, device=TORCH_DEVICE)
     cosmology = FlatLambdaCDM(name="cosmo")
-    cosmology.to(dtype=torch.float32)
+    cosmology.to(dtype=torch.float32, device=TORCH_DEVICE)
 
     # Parameters
     xs = [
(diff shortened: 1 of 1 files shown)


lsetiawan:
Running on a cluster, this PR currently fails with the following traceback:

```python
>>> import caustics
>>> caustics.test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.11/site-packages/caustics/tests.py", line 255, in test
    _test_simulator_runs()
  File "/opt/conda/lib/python3.11/site-packages/caustics/tests.py", line 56, in _test_simulator_runs
    assert torch.all(torch.isfinite(sim()))
                                    ^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/sims/simulator.py", line 32, in __call__
    return self.forward(packed_args, *rest_args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/sims/lens_source.py", line 217, in forward
    bx, by = self.lens.raytrace(*self.grid, z_s, params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/parametrized.py", line 576, in wrapped
    return method(self, *args, params=x, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/lenses/base.py", line 825, in raytrace
    ax, ay = self.reduced_deflection_angle(x, y, z_s, params, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/parametrized.py", line 576, in wrapped
    return method(self, *args, params=x, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/lenses/sie.py", line 129, in reduced_deflection_angle
    x, y = translate_rotate(x, y, x0, y0, phi)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/caustics/utils.py", line 56, in translate_rotate
    xt = x - x0
         ~~^~~~
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

src/caustics/tests.py
```diff
@@ -32,18 +37,22 @@ def _test_simulator_runs():
...
         lens_light=lenslight,
         psf=psf,
-        z_s=2.0,
+        z_s=z_s,
     )
 
+    # Send to device
+    sim = sim.to(TORCH_DEVICE)
+
     assert torch.all(torch.isfinite(sim()))
```
lsetiawan, line 56: Looks like the failure that I pointed out to traceback is on this line, where `sim()` is called.. I had thought that I mitigate this by calling `sim.to` on L54. I must be missing something else.

ConnorStoneAstro: You are right, the `.to` method should have handled this. I'll take a look

src/caustics/tests.py
```diff
@@ -10,14 +10,24 @@
...
-def _test_simulator_runs():
+TORCH_DEVICE = (
+    torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+)
+
+
+def _test_simulator_runs(device=TORCH_DEVICE):
     # Model
     cosmology = FlatLambdaCDM(name="cosmo")
+    z_l = torch.tensor(1.0, device=device)
```
ConnorStoneAstro, line 26: Setting `z_l` to a device shouldn't be necessary because you have `sim.to` later. Though it seems that isn't working, right?

src/caustics/tests.py
```diff
@@ -10,14 +10,24 @@
 
 __all__ = ["test"]
 
+# Pseudo device as mentioned in https://github.com/pytorch/pytorch/issues/61654
+# using this device, no actual computation is done, but will check
+# for where the tensors are located
+META_DEVICE = torch.device("meta")
```
ConnorStoneAstro, line 16: This is very cool! can you use this meta device in the tests here?

src/caustics/tests.py
```diff
@@ -10,14 +10,24 @@
...
 
-def _test_simulator_runs():
+TORCH_DEVICE = (
```
ConnorStoneAstro, line 18: I've usually seen it written as just `DEVICE` not `TORCH_DEVICE`. We could change that just to keep with convention.

src/caustics/tests.py
ConnorStoneAstro: I think I found the issue, I pushed it to dev. Can you try updating to the latest dev branch?

src/caustics/tests.py
lsetiawan: I can revert these changes to see if that works out.

src/caustics/tests.py
lsetiawan: Yea, I'm planning to use the meta device in a separate PR, since this doesn't really perform computation when used. I figured is a whole separate test of the software.. in this case, just to ensure that Tensors are placed in a single device before compute happens.

lsetiawan:
Woo 🚀 It looks like all is working properly with this simple test

```console
[landungs@g3073 landungs]$ apptainer run --nv caustics_dev-tests.sif /bin/bash
Apptainer> python
Python 3.11.7 | packaged by conda-forge | (main, Dec 23 2023, 14:43:09) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import caustics
>>> caustics.test()
all tests passed!
>>> from caustics.tests import DEVICE
>>> DEVICE
device(type='cuda')
>>> caustics.test(DEVICE)
all tests passed!
```

ConnorStoneAstro (approved):
Very Exciting!

PR #163, opened 2024-02-15
dependencies: move numba req to dev, lose levmarq_torch req

ConnorStoneAstro:


lsetiawan (commented):
The part about numba makes sense.. I think the previous issue was coming from lenstronomy.

PR #168, opened 2024-03-05
feat: Add properties to get x keys and order

lsetiawan:
### Overview

This PR extracts the x keys algorithm to a separate property so that user can get at it from `x_keys`... additionally, users can now call `x_order` to get the x keys in order so they can have information about an x tensor array.

### DIFF
--- a/src/caustics/parametrized.py
+++ b/src/caustics/parametrized.py
@@ -1,8 +1,9 @@
 # mypy: disable-error-code="var-annotated,index,type-arg"
 from collections import OrderedDict
 from math import prod
-from typing import Optional, Union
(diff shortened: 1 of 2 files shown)


lsetiawan:
> Just one question, can this handle an arbitrarily deep hierarchy? Like if I have a simulator with a multiplane lens, with a single plane lens, with an sie lens, will the x_keys and x_order collect that?

Hmm... great question! I'm not quite sure. I simply copied over what was already written in the code and extracted it out. Let me try it real quick and get back to you on that!

ConnorStoneAstro (approved):

PR #171, opened 2024-03-07
Streamlit: A visualization tool for Caustics

mjyb16:
I have created a beta version of the Caustics visualization tool in the "gui" folder. It was a lot of fun! There are a few things to note:

- For some reason, I have to rotate the imshow plots 90 deg to get them to face the right way (as far as I can tell from playing with the parameters). It is worth checking this just to make sure.

- The current lens+source configuration's redshift parameters don't change anything in the visualization, which I believe is what we expect from the reduced deflection angles and scale-invariant power laws, but I could be mistaken. 

- The shear components are interesting: they don't seem to be exactly orthogonal. One seems to be a vertical component, while the other seems to be a diagonal component.

However, these issues pertain more to the core code of Caustics than the visualization tool, so I anticipate some of them may become issues of their own. They should not impede the adoption of this simulator tool, since this tool should make debugging lens models easier. 

The only additional software required to use this tool is streamlit, which is a simple pip/conda install away. The tool can be run using the following command: streamlit run "path_to/caustics/gui/streamlit_app.py"

Moving forward, we can add more models to the menu on the left below the Caustics logo, as well as hosting + adding to the website.


ConnorStoneAstro:
This is really amazing! There are a few simple changes that could be helpful:
- No need to include the redshift sliders, they dont do anything
- No need to include the x0 and y0 for the shear (we may even remove those parameters from caustics entirely)
- No need to include the sersic intensity
- The model can be evaluated at lower resolution, should help it run faster
- The caustic should appear in the source plane image, but in the lens plane you should show the critical curve
- Caustics should be computed in the same fov as the image so they don't get cut off near the edges
- Rotations can go from 0 to pi instead of 0 to 2pi since everything is symmetric
- Adding the rest of the lens models (but you knew that)!

Great stuff!

ConnorStoneAstro:
Also, 

- if you could use the inferno cmap, I just think it looks cool :P 
- There is an updated script to compute contours, that will probably work better. Check out the new "Visualize Caustics" tutorial
- What is the `streamlit_app_basic` for? Should it be removed?

mjyb16:
I also removed the "basic" app, that was mostly for my experimentation (to keep something stable while I messed around with the main simulator).

ConnorStoneAstro (approved):
Looks Great To Me :)

PR #173, opened 2024-03-07
Adding example notebooks

ConnorStoneAstro:
Here I show some near real world analysis routines with mocked data and resulting analysis.


mjyb16:
Needs a couple more markdown cells explaining what is going on in each plot, eg clarifying whether the vertical lines in the corner plot are the ground truth. Also needs a check on the second row of Uncertainties: why is the source row showing lensed values?

PR #178, opened 2024-03-09
Allow selecting different lenses in streamlit app

ConnorStoneAstro:
Users can now select as many lenses as they like from a menu and stack them together. The parameters are auto-built from the slider positions.


ConnorStoneAstro:
This PR is no longer needed as the streamlit app is moving to a separate repo.

PR #188, opened 2024-03-13
Fix mpmath to non beta version

ConnorStoneAstro:


requirements.txt
```diff
@@ -1,6 +1,7 @@
 astropy>=5.2.1,<6.0.0
 graphviz==0.20.1
 h5py>=3.8.0
+mpmath<1.4
```
uwcdc, line 4: Do you want to set it for less than 1.4 for the latest version that we know works?
```suggestion
mpmath==1.3.0
```

uwcdc: @ConnorStoneAstro  you may want to test out compatibility with other versions of `mpmath` in the future.

ConnorStoneAstro: @uwcdc setting `mpmath<1.4` was @lsetiawan 's suggestion yesterday. I think because we want to benefit from bug patches, but we don't want any api changes.

The error we were getting was that a function didn't exist, which should only happen with minor/major version updates, not patch.

What do you think?

uwcdc: @ConnorStoneAstro I understand where you are coming from. In that case, let's bracket the compatibility to include lower minor releases but exclude greater than 1.4.
```suggestion
mpmath=<1.3, <1.4.0
```

ConnorStoneAstro: Ah, great idea! That's what I'll do in the morning

ConnorStoneAstro:
Closes #189

uwcdc (approved):

PR #190, opened 2024-03-13
Streamlit: improved caustic tracing

mjyb16:
Caustic tracing is sped up by 40% (and accuracy is improved), no longer requires matplotlib (now with scikit-image), and is now optional in the simulator.


ConnorStoneAstro:
Now in it's own repo: caustics-webapp

PR #191, opened 2024-03-14
feat: Adding `caustics.func`

ConnorStoneAstro:
This adds a new interface level for caustics with fully functional base functions for the caustics OOP layer. This allows for more precise access to functions without building objects.

To access the reduced deflection angle calculations for an SIE it is now possible to directly call:
`caustics.func.reduced_deflection_angle_sie`

instead of building a `caustics.SIE` object and then calling the reduced deflection angle method. This allows for one to build simulators in a very granular way. However, the OOP caustics interface is still what we expect to be the primary interface for most users, the func design is for very advanced analyses or unusual design requirements.


AlexandreAdam:
I haven't gone through everything yet but I love the name convention, i.e. put the profile at the end of the name and the quantity of interest first. 

This is a rather extensive PR. Can you add some notes or mention notable changes to the code that were in the classes before? I'll start to review the lenses during the week.

Also, I'm curious to know how you handle the parametrization, following the discussion in #196 .

ConnorStoneAstro:
@AlexandreAdam Glad you like the convention! :) 

My plan for this PR is to change as little as possible in how the code operates, I'm just copy-pasting. So if you look at what's happened to one lens you get the idea for everything. That said, there are some decisions for e.g. NFW and TNFW which I'll post in more detail once I have a functioning version of this put together.

I plan to copy the classes as written here. So #196 will be entirely independent.

ConnorStoneAstro:
Ok I think the `func` design is ready to go, sorry this PR is basically unreadable in git... 

Most of the `func` methods are directly copied from before so there is no change in behavior. While going through everything I identified some minor bits of redundant code and cleaned them up, but those were very straightforward. 

Some places required a bit more thought though. For the NFW we run into the issue that PyTorch cannot currently evaluate `where` without propogating nans from the `False` branch in autodiff (internally `where` just multiplies the `False` values by zero, which doesn't eliminate nans in autodiff, although it inexplicably works in the forward pass). So the user must decide if they want autodiff or if they want vmap. I implimented this by allowing the user to pass a function as an argument, so they can pick the differentiable or the batchable one. For example for the `physical_deflection_angle_nfw` function there is an argument `_h` where the user can select `_h_differentiable_nfw` or `_h_batchable_nfw` which are also defined in `func`, the default function is `_h_differentiable_nfw`. A similar setup is used for `TNFW`.

For the `TNFW` there is a normalization value `M0` which can be computed using the "mass" of an NFW which is an arbitrary integral out to r200 for an NFW (not a TNFW), or it can be set to the total mass of the TNFW. The second one is way more meaningful, but often people will want to switch between NFW and TNFW in which case the first is kind of better. To allow for flexibility I let the user provide the `M0` value as an argument, then they can use either the `M0_scalemass_tnfw` or the `M0_totalmass_tnfw ` functions themselves.

Some very trivial things like the convergence of a mass sheet, I do not create a `func` version. Though we may want to add them at some point just for completeness.

I absorb a bunch of the helper functions from before into the `func` module, so this does make the classes a bit cleaner which is nice.

Also I added references and equation numbers for every function.

src/caustics/lenses/func/base.py
```diff
@@ -0,0 +1,163 @@
...
+        *Unit: arcsec*
+
+    y_component: Tensor
+        Physical deflection Angle in the y-direction.
+
+        *Unit: arcsec*
+
+    """
+
+    return ((d_s / d_ls) * ax, (d_s / d_ls) * ay)
```
AlexandreAdam, line 120: One extra parenthesis in this expression that is not needed

src/caustics/lenses/func/base.py
```diff
@@ -0,0 +1,163 @@
...
+        *Unit: arcsec*
+
+    y_component: Tensor
+        Reduced deflection Angle in the y-direction.
+
+        *Unit: arcsec*
+
+    """
+
+    return ((d_ls / d_s) * ax, (d_ls / d_s) * ay)
```
AlexandreAdam, line 163: Same

AlexandreAdam (approved):
This look good to me, I only have one minor comment that can be ignored. Good job!

PR #203, opened 2024-04-05
feat: Lens converter classes

### LINKED ISSUE #202: Add basic lens quantity converters
We already have pixelated convergence, we should add a pixelated potential method as well. There are other transformations that can be made. If we have an axisymmetric model with a cumulative mass function, or a radial potential function it is possible to compute the deflection field. It is also possible to turn axisymmetric models into elliptical models with a slight substitution (meneghetti notes 3.87). Thus we could make it so a user can just input any function they want for the radial density and get a deflection field.

ConnorStoneAstro:
This PR adds lens models which convert various quantities. For example, given the enclosed mass profile it can compute a deflection field.

Closes #202


AlexandreAdam (approved):
This looks good. Consistency test look good to me as well. This is a nice natural addition

PR #206, opened 2024-04-24
Turned pydantic into a weak dependancy

AlexandreAdam:
# Hot Fix

Turning `pydantic` into a weak dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest pytest-cov torch wheel
+          pip install pytest pytest-cov torch wheel pydantic
 
       # We only want to install this on one run, because otherwise we'll have
       # duplicate annotations.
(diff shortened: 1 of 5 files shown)


ConnorStoneAstro:
Since pydantic isnt a req anymore, you'll need to modify the github action scripts to manually install pydantic so that the tests run

AlexandreAdam:
I've added pydantic to the CI workflow, I'll need to iterate with the automatic testing because I'm not sure if it's enough

AlexandreAdam:
Good catch, although to address this I need to change a few more things. I need to make sure the file with pydantic import is not part of the Caustics root import, and is only imported in the function that requires it, which is in the simulator I believe

AlexandreAdam:
Correct me if I am wrong, but I think the approach I took solve the problem. The only thing exposed to the root level imports is build_simulator, so I made sure all the import statements related to Pydantic were used only inside the function definition. As such, Pydantic is now made a weak dependency. Caustics should attempt to import it only if build_simulator is called

ConnorStoneAstro (approved):
Looks good to me! the imports should only happen when calling build simulator

Issue #139, opened 2023-12-28
Multipole expansions


mjyb16:
Is there a plan for adding multipole expansion parameters to lenses? I don't see any mentions of multipole capability at the moment in the codebase or the docs.

ConnorStoneAstro:
Hmmm, could you describe more of what you are looking for here? Is there a reference where this is done? For a lot of lenses they are quite simple to evaluate and a multipole expansion would probably just increase the computational graph. An exception might be the `NFW` and `TNFW` lenses which are a bit of a pain right now. Or are you thinking of a new lens class that is just a multipole expansion up to a certain order and can take arbitrary values?

AlexandreAdam:
Multipole should be an additional lens class. The external convergence is a very special case of a multipole extension, since it is the zeroth moment of the multipole expansion of the external convergence. 

A good reference on the subject is [Schneider & Bartelmann (1997)](https://articles.adsabs.harvard.edu/pdf/1997MNRAS.286..696S), but one can find also a more modern treatment in [Bernstein & Nakajima](https://iopscience.iop.org/article/10.1088/0004-637X/693/2/1508/pdf). 

Note that Yashar's [SDP.81 paper](https://arxiv.org/abs/1601.01388) found that multiples were very important in the detection of substructure. He gives a simple definition of the multiple expansion of the convergence in polar coordinates
```math
\kappa(r, \theta) = \left(\frac{r}{r_s}\right)^\alpha [A_m \cos (m \phi) + B_m \sin(m \phi)]
```
In his paper he describes that the logarithmic slope $\alpha$ is shared with his main lens parameters and he chooses $r_s$ based on the size of his lens (?).

ConnorStoneAstro:
closed by #215

PR #215, opened 2024-05-29
feat: add multipole lens profile

andreasfilipp:
adresses issue #139 

coauthor: @RonanLegin 

**Multipole Gravitational Lensing Module**

This Python module, part of a larger gravitational lensing library, provides functionality to simulate and analyze multipole gravitational lensing effects. It is designed to integrate seamlessly with PyTorch and offers detailed implementations for calculating deflection angles, lensing potentials, and convergence based on multipole expansions.

**Features**
Gravitational Lensing Calculations: Compute deflection angles, potentials, and convergence for multipole gravitational lenses.

**Documentation**
Refer to the docstrings within the code for detailed usage and examples. Equations used for the calculations are referenced from "Xu et al. 2014" (https://arxiv.org/pdf/1307.4220).

### DIFF
--- a/src/caustics/lenses/func/multipole.py
+++ b/src/caustics/lenses/func/multipole.py
@@ -37,6 +37,8 @@ def reduced_deflection_angle_multipole(x0, y0, m, a_m, phi_m, x, y):
 
     phi = torch.arctan2(y, x).reshape(1, -1)
     m = m.reshape(-1, 1)
+    a_m = a_m.reshape(-1, 1)
+    phi_m = phi_m.reshape(-1, 1)
     ax = torch.cos(phi) * a_m / (1 - m**2) * torch.cos(m * (phi - phi_m)) + torch.sin(
         phi
     ) * m * a_m / (1 - m**2) * torch.sin(m * (phi - phi_m))
@@ -83,9 +85,11 @@ def potential_multipole(x0, y0, m, a_m, phi_m, x, y):
 
     """
     x, y = translate_rotate(x, y, x0, y0)
-    r = torch.sqrt(x**2 + y**2)
+    r = torch.sqrt(x**2 + y**2).reshape(1, -1)
     phi = torch.arctan2(y, x).reshape(1, -1)
     m = m.reshape(-1, 1)
+    a_m = a_m.reshape(-1, 1)
+    phi_m = phi_m.reshape(-1, 1)
 
     potential = r * a_m / (1 - m**2) * torch.cos(m * (phi - phi_m))
     potential = potential.sum(dim=0).reshape(x.shape)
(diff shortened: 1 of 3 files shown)


andreasfilipp:
the convergence test for some reason does not pass. I do not know why

src/caustics/lenses/multipole.py
```diff
@@ -0,0 +1,228 @@
...
+
+        y_component: Tensor
+            Deflection Angle
+
+            *Unit: arcsec*
+
+        Equation (B11) and (B12) https://arxiv.org/pdf/1307.4220, Xu et al. 2014
+
+        """
+        return func.reduced_deflection_angle_multipole(x0, y0, m, a_m, phi_m, x, y)
```
ConnorStoneAstro, line 126: Could you make this go over a loop in the `self.m` values and add the contributions? Could even vmap it.

src/caustics/lenses/multipole.py
```diff
@@ -0,0 +1,228 @@
...
+        -------
+        Tensor
+            The lensing potential.
+
+            *Unit: arcsec^2*
+
+        Equation (B11) and (B3) https://arxiv.org/pdf/1307.4220, Xu et al. 2014
+
+        """
+        return func.potential_multipole(x0, y0, m, a_m, phi_m, x, y)
```
ConnorStoneAstro, line 177: same loop in `self.m`

src/caustics/lenses/multipole.py
```diff
@@ -0,0 +1,228 @@
...
+        -------
+        Tensor
+            The projected mass density.
+
+            *Unit: unitless*
+
+        Equation (B10) and (B3) https://arxiv.org/pdf/1307.4220, Xu et al. 2014
+
+        """
+        return func.convergence_multipole(x0, y0, m, a_m, phi_m, x, y)
```
ConnorStoneAstro, line 228: same loop in `self.m`

src/caustics/models/utils.py
```diff
@@ -48,7 +48,7 @@ def _get_kwargs_field_definitions(
         if k != "name":
             anno = v.annotation
             dtype = anno.__origin__
-            cls_param = ClassParam(*anno.__metadata__)
+            cls_param = ClassParam(*anno.__metadata__)  # gooooo
```
ConnorStoneAstro, line 51:
```suggestion
            cls_param = ClassParam(*anno.__metadata__)
```

ConnorStoneAstro (changes requested):
Some minor changes, and one bigger change. Could you make it so that the user passes `m` as a tuple of which multipole orders they want? Then the individual functions can loop over that tuple and add the contributions from each multipole order.

Also can you add the multipole class to the `test_lens_potential.py` unit test? It doesn't look like its in there although I thought I saw you running that test

PR #217, opened 2024-06-08
feat: Added pixelated cube with time dimension

ConnorStoneAstro:
Same as `Pixelated` source, except it is now indexed by x, y, t to allow for time varying sources


AlexandreAdam (approved):
This looks good to me

PR #218, opened 2024-06-08
feat: Add microlensing flux simulator

ConnorStoneAstro:
New simulator class evaluates a number of raytraces and then produces an estimate of the flux, includes an estimate of the uncertainty on the flux.


ConnorStoneAstro:
@AlexandreAdam I have now add the `Literal` type annotation for the method argument

PR #220, opened 2024-06-13
refactor: Fix magnification input shape to allow any shape

### LINKED ISSUE #219: `get_magnification` requires image input, but should accept any shape tensor
Currently it is required when calling `lens.magnification(x,y,z_s)` for the `x,y` arguments to have exactly two dimensions. We can fix this to allow any shape input.

ConnorStoneAstro:
Closes #219


AlexandreAdam (changes requested):
Do we have a test that can catch the problem associated with this change for future PRs?

If not, we should add one as part of this PR.

ConnorStoneAstro:
Ok, added a test. While adding it I discovered that a special case was needed if the user inputs single values since this returns an empty shape. Now should be able to handle coordinates that are scalars, or any shaped tensor.

ConnorStoneAstro:
Actually, I figured out a way to do it without the if statement. Should be good now @AlexandreAdam

PR #227, opened 2024-06-28
Interface cleanup

AlexandreAdam:
# List of changes
- Removed `n_pix` from the initialization of `PixelatedConvergence`, This is now inferred either form shape or the image itself
- `meshgrid` now has signature `meshgrid(pixelscale, pixels)` with optional signature `meshgrid(pixelscale, nx, ny)` in case grid is not a square. 
- Cleaned up the `get_*`


src/caustics/lenses/pixelated_convergence.py
```diff
@@ -145,7 +140,10 @@ def __init__(
         self.add_param("y0", y0)
         self.add_param("convergence_map", convergence_map, shape)
 
-        self.n_pix = n_pix
+        if convergence_map is not None:
```
ConnorStoneAstro, line 143: Should probably throw an error if convergence_map and shape are both missing

ConnorStoneAstro (commented):
One general comment, perhaps we should change `meshgrid` to something like `meshpixels` or `pixelmesh` since numpy and pytorch already have `meshgrid` it is a bit confusing to overload the meaning of this name.

src/caustics/lenses/pixelated_convergence.py
AlexandreAdam: Hmm, this error is caught above in line 132-138 roughly.

AlexandreAdam:
I disagree, we don't want to introduce more concepts. Meshgrid will be intuitive for anybody that used torch/numpy etc. if they know it. Plus, we don't expose this in the interface. It's more of a utils.

ConnorStoneAstro (approved):

PR #228, opened 2024-06-28
ci: fix codecov token

ConnorStoneAstro:


AlexandreAdam:
I'm fine with deleting these. They are outside the scope of caustics for sure.

AlexandreAdam (approved):

PR #236, opened 2024-07-05
docs: try to get github link to appear on website

ConnorStoneAstro:


ConnorStoneAstro:
Doesn't seem to work even with every option enabled. We can revisit this another time

PR #241, opened 2024-07-16
docs: try to make window inline

ConnorStoneAstro:
Just testing if we can get the streamlit app to render in the website instead of by a link


ConnorStoneAstro:
Misha: "if it doesn't show up in your window, click this link"

ConnorStoneAstro:
Maybe minimize sidebar by default?

mjyb16 (approved):
Looks good!

PR #245, opened 2024-07-26
refactor: move quad_level to init

ConnorStoneAstro:
Moving the argument `quad_level` to the init of `LensSource`. I think this would also be a good PR to refactor the internals of `LensSource` since it has gotten to be quite nested. Let's use this to discuss how it should look.

### DIFF
--- a/src/caustics/sims/lens_source.py
+++ b/src/caustics/sims/lens_source.py
@@ -74,12 +74,6 @@ class LensSource(Simulator):
         sample the image at a higher resolution, then integrate the image back
         to the original resolution. This is useful for high accuracy integration
         of the image, but may increase memory usage and runtime.
-    psf_pad: Boolean(default True)
-        If convolving the PSF it is important to sample the model in a larger
-        FOV equal to half the PSF size in order to account for light that
-        scatters from outside the requested FOV inwards. Internally this padding
-        will be added before sampling, then cropped off before returning the
-        final image to the user.
     z_s: optional
         redshift of the source
     name: string (default "sim")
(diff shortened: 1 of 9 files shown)


src/caustics/sims/lens_source.py
```diff
@@ -91,18 +120,21 @@ def __init__(
...
-        psf_pad: Annotated[bool, "Flag to apply padding to psf"] = True,
+        quad_level: Annotated[Optional[int], "sub pixel integration resolution"] = None,
         psf_mode: Annotated[
             Literal["fft", "conv2d"], "Mode for convolving psf"
         ] = "fft",
+        psf_shape: Annotated[Optional[tuple[int, ...]], "The shape of the psf"] = None,
         z_s: Annotated[
             Optional[Union[Tensor, float]], "Redshift of the source", True
         ] = None,
+        psf: Annotated[
```
AlexandreAdam, line 135: Would make it None, easier to handle as a default.

src/caustics/sims/lens_source.py
```diff
@@ -229,75 +330,63 @@ def forward(
...
-        z_s, x0, y0 = self.unpack(params)
+        z_s, psf, x0, y0 = self.unpack(params)
 
         # Automatically turn off light for missing objects
         if self.source is None:
             source_light = False
         if self.lens_light is None:
             lens_light = False
-        if self.psf is None:
+        if psf.shape == (1, 1):
```
AlexandreAdam, line 340: If we use a kernel of 1 means we always convolve with the PSF, do we still need this?

src/caustics/sims/lens_source.py
ConnorStoneAstro: None as a default means dynamic, so user specified. I wanted the default to not require a PSF. also there is a natural "null psf" which is the 1x1 kernel since that must always be the identity because `psf = psf / psf.sum()`.

src/caustics/sims/lens_source.py
ConnorStoneAstro: I think this is still useful. Convolving with a 1x1 kernel using the default fft method involves an fft and inverse fft. This unecessarily limits the dynamic range for a null operation. So just adding a little check to skip it saves someone else a headache.

AlexandreAdam: Yeah just realized there is the fft, though couldn’t we do something similar to the Pixelated Convergence where we decide which method to use at init or with the property trigger?

AlexandreAdam: Basically we would only use fft if the shape of the PSF exceed something like 32x32

ConnorStoneAstro: I made it so the user can choose the method with psf_mode. I feel like an automatic selector might cause unexpected behavior

AlexandreAdam (approved):
Excellent!

PR #251, opened 2024-08-29
Add MH MCMC example

ConnorStoneAstro:
Creates a notebook like the NUTS example which uses metropolis hastings MCMC for comparison.


AlexandreAdam:
Ah maybe just one comment. To test if the chain mix well,
you plot the chain such that we can see them individually by adjusting the height each chain to be different.

I prefer the test where each chain is plotted on top of each other, each with a different color. This plot makes it easy to see if the chains have mixed because the colors of each chain will blend.  It's going to be hard to pick out one chain over another.

If the chains have not mixed then you'll be able to pick some of the chains and really tell that they are from a different stochastic process.

It's easy to do with a simple for loop
```
B = # number of chains 
for b in range(B):
     plt.plot(chain[b])
 ```
You can also do some pretty stuff if you rotate the colors from a perceptuably uniform cmap (https://colorcet.holoviz.org/user_guide/index.html).  And then you can do something really cool
```
import colorcet as cc

cmap = cc.cm.glasbey
for b in range(B):
    color = cmap(b % cmap.N) # Cycle through
    plt.plot(chain[b], lw=2, color=color)
```
Screenshot from 2024-08-31 15-36-21

I typically use this setup
```
import scienceplots # pip install SciencePlots
import colorcet as cc # pip install colorcet
import pylab
from tqdm import tqdm

plt.style.use('science')
params = {
         'axes.labelsize': 18,
         'axes.titlesize': 30,
         'ytick.labelsize' :20,
         'xtick.labelsize' :20,
         'legend.fontsize': 16,
         'xtick.major.size': 8,
         'xtick.minor.size': 4,
         'ytick.major.size': 8,
         'ytick.minor.size': 4,
         }
pylab.rcParams.update(params)
```
But it's really up to you, this is more of a suggestion than a request for change.

PR #259, opened 2024-09-16
Microlensing

NolanSmyth:
Adding minimal microlensing simulator functionality and model fitting example


src/caustics/light/point_source.py
```diff
@@ -0,0 +1,163 @@
...
+
+from .base import Source, NameType
+from ..parametrized import unpack
+from ..packed import Packed
+from . import func
+
+__all__ = ("PointSource",)
+
+
+class PointSource(Source):
```
ConnorStoneAstro, line 14: Point source might not be the right name for this class. A point source should have a delta function for its brightness, which isn't very practical. Perhaps a more direct name for this class like `StarSource` since that is what it is representing and it includes limb darkening?

src/caustics/light/point_source.py
```diff
@@ -0,0 +1,163 @@
...
+            Optional[Union[Tensor, float]],
+            "The intensity at the effective radius",
+            True,
+        ] = None,
+        gamma: Annotated[
+            Optional[Union[Tensor, float]],
+            "The linear limb darkening coefficient",
+            True,
+        ] = None,
+        s: Annotated[float, "A small constant for numerical stability"] = 0.0,
```
ConnorStoneAstro, line 75: I don't think `s` is needed here for numerical stability, so it can just be left out

src/caustics/light/func/point_source.py
```diff
@@ -0,0 +1,17 @@
+from torch import where
+
+from ...utils import translate_rotate
+
+
+def brightness_point(x0, y0, theta_s, Ie, x, y, gamma=1.0, s=0.0):
+
+    x, y = translate_rotate(
+        x, y, x0, y0, phi=None
```
ConnorStoneAstro, line 9: I believe phi is None by default so you can just not include it

src/caustics/light/point_source.py
```diff
@@ -0,0 +1,163 @@
...
+
+class PointSource(Source):
+    """
+    `Point` is a subclass of the abstract class `Source`.
+    It represents a point source in a gravitational lensing system.
+
+    The Point profile is meant to describe individual light sources
+    whose angular extent is much less than the Einstein radius of the lens.
+
+    Attributes
```
ConnorStoneAstro, line 22: need to include all the parameters

src/caustics/light/func/point_source.py
```diff
@@ -0,0 +1,17 @@
+from torch import where
+
+from ...utils import translate_rotate
+
+
+def brightness_point(x0, y0, theta_s, Ie, x, y, gamma=1.0, s=0.0):
```
ConnorStoneAstro, line 6: I don't think a default value is needed for `gamma` and if there is a default value, I would set it to `0.0` for no limb darkening to make the default behaviour more like one would expect. Unless `1.0` is some special value for limb darkening.

ConnorStoneAstro (changes requested):

ConnorStoneAstro:
I've changed the PR to point to `dev` that's where we develop updates before moving them to `main`

src/caustics/light/point_source.py
NolanSmyth: Good call. I was originally planning to implement a point source and then add a finite extent, but this isn't realistic. I renamed all relevant classes/functions to use "star" instead of "point"

src/caustics/light/func/point_source.py
NolanSmyth: My mistake. I meant to set gamma to 0.0 for no limb darkening by default, which I think makes sense. I believe most users would assume no limb darkening as a starting point; it is often treated as a higher-order effect and the coefficient can take a wide range of values.

NolanSmyth:
@ConnorStoneAstro Do you see anything else that needs to be fixed, or good to go?

ConnorStoneAstro:
@NolanSmyth it looks like there are still merge conflicts. Also the tutorial wasn't added to the TOC correctly, I think the capitalization is wrong so it's not getting picked up.

ConnorStoneAstro:
Ah, looks like you'll need unit tests as well. Just something simple will do where you test it a few parameters to make sure you don't get a nan anywhere

docs/source/_toc.yml
```diff
@@ -21,6 +21,7 @@ chapters:
       - file: tutorials/InvertLensEquation
       - file: tutorials/Parameters
       - file: tutorials/Simulators
+      - file: tutorials/Microlensing
```
ConnorStoneAstro, line 24: The TOC still has the filename wrong so it's not rendering into the docs

```suggestion
      - file: tutorials/microlensing
```

NolanSmyth: There was a weird issue where the capitalization wasn't tracked (it was capitalized correctly on my Macbook)
image

As a quick work around, did the following: rename the file, commit, rename it back (with proper capitalization) and commit again. It looks like that fixed the issue!

docs/source/tutorials/Microlensing.ipynb
```diff
@@ -0,0 +1,318 @@
...
+    "        th_ein,  # th_ein\n",
+    "        0.0,  # x0\n",
+    "        0.0,  # y0\n",
+    "        th_ein * rho,  # theta_s\n",
+    "        5.0,  # Ie,\n",
+    "        gamma,  # gamma\n",
+    "    ]\n",
+    ")\n",
+    "\n",
+    "# Create the input tensor for the simulator. In this simple example, everything is static except for the lens position\n",
```
ConnorStoneAstro, line 203: If everything except the lens position `x0` is static then you can set the parameters as fixed using calls like `sim.lens.th_ein = th_ein` and then the `x` you input to the simulator only needs to include the `x0` value

docs/source/tutorials/Microlensing.ipynb
```diff
@@ -0,0 +1,318 @@
...
+    "images, magnifications = torch.vmap(sim)(newx)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create animation\n",
```
ConnorStoneAstro, line 233: You can add a tag to this cell `hide-input` so that users will only see the plot and not all the plt commands, it makes for a cleaner look.

.pre-commit-config.yaml
```diff
@@ -69,7 +69,13 @@ repos:
     rev: "v2.3.0"
     hooks:
       - id: codespell
-        args: ["--write-changes", "--ignore-words", ".codespell-whitelist"]
+        args:
+          [
+            "--write-changes",
+            "--ignore-words",
+            ".codespell-whitelist",
+            "--ignore-words-list=tE",
```
ConnorStoneAstro, line 77: You can add `tE` to the codespell-whitelist. It should be `.codespell-whitelist` in the main directory of caustics

docs/source/examples/Simulated_Event.html
ConnorStoneAstro, line 1: I'm not sure what this file is doing, I don't see it appear in the examples in the docs

docs/source/tutorials/assets/microlensing_dat.npz
ConnorStoneAstro, line 1: I think this file isn't needed

ConnorStoneAstro (changes requested):
This is so cool! I love the animation, we'll have to do more of those! I just have a few comments which should be quick to change.

ConnorStoneAstro (approved):
I love it!

PR #264, opened 2024-09-26
Improve forward raytrace to better find all images

ConnorStoneAstro:


src/caustics/lenses/func/base.py
```diff
@@ -4,7 +4,106 @@
...
+        # Compare current triangle with all triangles in the unique list
+        if i == 0 or not batch_triangle_equals(p[i], unique_triangles).any():
+            unique_triangles = torch.cat((unique_triangles, p[i].unsqueeze(0)), dim=0)
+
+    return unique_triangles
+
+
+def forward_raytrace_rootfind(ix, iy, bx, by, raytrace):
     """
     Perform a forward ray-tracing operation which maps from the source plane to the image plane.
```
NolanSmyth, line 108: Could include docstrings for all parameters

docs/source/tutorials/InvertLensEquation.ipynb
```diff
@@ -219,10 +276,290 @@
...
+    "ax.set_ylim([-fov / 1.9, fov / 1.9])\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b4c5d47b",
+   "metadata": {},
+   "source": [
+    "Let's now look at a more clever algorithm. We will map triangles in the image plane to triangles in the source plane, we may then explore recursively, any triangles which enclose the desired source point. Due to the non-linearity of the gravitational lensing transformation, we will also search the neighbor of any triangle that seems to have found an image position. First we highlight in green, any triangles which contain the source point, then expand to all their neighbors."
```
NolanSmyth, line 370: In the 2nd iteration, it looks like there are many more triangles included than just those that enclose the source (green triangles) and their nearest neighbors. Is that right? If so, I don't see why one would want to search beyond the immediate neighbors.
image

docs/source/tutorials/InvertLensEquation.ipynb
```diff
@@ -219,10 +276,290 @@
...
+    "ax.set_axis_off()\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "61fdd482",
+   "metadata": {},
+   "source": [
+    "The process repeats until the triangles have converged to a very small area, at which point we then run a root finding algorithm to get the final points."
```
NolanSmyth, line 484: If I overlay the demagnified image found using the full `forward_raytrace` in the first section with the final images produced here, it looks as though they don't overlap. At first, I thought this was because the recursive iteration was stopped when it was still coarse. But it looks like you're running `forward_raytrace_rootfind` on the solutions to get the true points, so I'm not sure why they appear so offset. 
image

NolanSmyth (changes requested):
Very cool algorithm! Just had a couple questions about the behavior in the tutorial notebook.

docs/source/tutorials/InvertLensEquation.ipynb
ConnorStoneAstro: Ah so the neighbors are created with the larger triangles before upsampling them, so you would want to look at the figure before this one to see where the neighbors should be. Either way, yes this is searching a pretty large area, but the lens images can be quite elusive if we are unlucky, so this is just being cautious.

docs/source/tutorials/InvertLensEquation.ipynb
ConnorStoneAstro: Yeah, I'll have to look into that. My guess is that the root finder is getting stuck and ends up just returning the input point. But the full forward_raytrace algorithm does go much deeper in the recursion, so I think that one is very accurate. Still, I'll see what's going on with the root finder.

ConnorStoneAstro:
@NolanSmyth It should be ready now, what do you think?

NolanSmyth (approved):

PR #275, opened 2024-10-23
refactor: Change Parametrized system over to caskade

ConnorStoneAstro:
`Parametrized` has been extracted from caustics and improved into the new `caskade` system. This PR ports caustics over to the new functionality.


ConnorStoneAstro:
Remember to change order of lens source so that z_s goes before the modules

src/caustics/lenses/batchedplane.py
```diff
@@ -0,0 +1,198 @@
+import torch
+from torch import Tensor
+from caskade import forward
+
+from .base import ThinLens, CosmologyType, NameType, ZLType
+
+__all__ = ("BatchedPlane",)
+
+
+class BatchedPlane(ThinLens):
```
AlexandreAdam, line 10: That's awsome!

src/caustics/lenses/batchedplane.py
```diff
@@ -0,0 +1,198 @@
...
+
+            *Unit: unitless*
+
+        """
+        # Collect the dynamic parameters to vmap over
+        params = dict((p.name, p.value) for p in self.lens.local_dynamic_params)
+        batchdims = dict(
+            (p.name, -(len(p.shape) + 1)) for p in self.lens.local_dynamic_params
+        )
+        convergence = torch.vmap(
```
AlexandreAdam, line 144: I worry that this operation can consume a lot of memory at its peak. In practice, when drawing convergence, we often deal with large population. If not careful, the GPU memory will be entirely consumed by copies of the plane to draw the halo on. 

In practice, I prefer to use a mix of vmap and for loop to manage the peak memory usage of this operation. This requires defining a batch_size in the init and doing something like
```python
total = params.shape[0]
convergence = 0
for _ in range(total//batch_size):
    convergence += vmap(self.lens.convergence(...)).sum(0)
return convergence
```

AlexandreAdam (commented):
Awsome stuff!

src/caustics/lenses/batchedplane.py
ConnorStoneAstro: I think I could just use the `chunk_size` parameter in vmap, but yeah that's a good point I can expose that option to the user

PR #301, opened 2024-12-19
fix: handle q=1 for sie

ConnorStoneAstro:


mjyb16 (approved):
I trust you on the numerical stability issues. I will be testing this on real data soon so if something crops up we can patch.

PR #305, opened 2024-12-20
refactor: Change z_s from argument to caskade param

ConnorStoneAstro:
Now that caskade handles linking parameters, it is possible to make z_s a param and simply link the value where appropriate. 

This update means less confusion for users about when/why z_s should be passed. It is intuitive that a lens object may need z_s, some functions however don't need it yet were previously requiring it.


mjyb16 (approved):
Looked through the multisource in particular, looks much cleaner now with all the override functionality hidden. As far as I can tell, we have standardized things now where z_s is a lens parameter as opposed to a source parameter, which can be a bit confusing but as long as we are clear on it this is fine. I think this is good to go!

Issue #242, opened 2024-07-19
Fix circular logic in Thin Lens deflection angle calculation for multiplane code


AlexandreAdam:
In the ThinLens class we have method
```
    def physical_deflection_angle(
        self,
        x: Tensor,
        y: Tensor,
        z_s: Tensor,
        *args,
        params: Optional["Packed"] = None,
        z_l: Optional[Tensor] = None,
        **kwargs,
    ) -> tuple[Tensor, Tensor]:
        """
        Computes the physical deflection angle immediately after passing through this lens's plane.

        Parameters
        ----------
        x: Tensor
            Tensor of x coordinates in the lens plane.

            *Unit: arcsec*

        y: Tensor
            Tensor of y coordinates in the lens plane.

            *Unit: arcsec*

        z_s: Tensor
            Tensor of source redshifts.

            *Unit: unitless*

        params: (Packed, optional)
            Dynamic parameter container for the lens model. Defaults to None.

        Returns
        -------
        x_component: Tensor
            Deflection Angle in x-direction.

            *Unit: arcsec*

        y_component: Tensor
            Deflection Angle in y-direction.

            *Unit: arcsec*

        """
        d_s = self.cosmology.angular_diameter_distance(z_s, params)
        d_ls = self.cosmology.angular_diameter_distance_z1z2(z_l, z_s, params)
        deflection_angle_x, deflection_angle_y = self.reduced_deflection_angle(
            x, y, z_s, params
        )
        return func.physical_from_reduced_deflection_angle(
            deflection_angle_x, deflection_angle_y, d_s, d_ls
        )
```

We also have method
```
    def reduced_deflection_angle(
        self,
        x: Tensor,
        y: Tensor,
        z_s: Tensor,
        *args,
        params: Optional["Packed"] = None,
        z_l: Optional[Tensor] = None,
        **kwargs,
    ) -> tuple[Tensor, Tensor]:
        """
        Computes the reduced deflection angle of the lens at given coordinates [arcsec].

        Parameters
        ----------
        x: Tensor
            Tensor of x coordinates in the lens plane.

            *Unit: arcsec*

        y: Tensor
            Tensor of y coordinates in the lens plane.

            *Unit: arcsec*

        z_s: Tensor
            Tensor of source redshifts.

            *Unit: unitless*

        params: (Packed, optional)
            Dynamic parameter container for the lens model. Defaults to None.

        Returns
        --------
        x_component: Tensor
            Deflection Angle in the x-direction.

[...]

PR #306, opened 2024-12-20
refactor: better use of autograd for base lens quantities

ConnorStoneAstro:
Now one need only define a lens potential and the deflection angles will be computed from it's gradient, and the convergence will be computed from the hessian. Also, implemented better version of the jacobian of the deflection angles.


ConnorStoneAstro:
This may also address #242

AlexandreAdam (commented):
NFW and tNFW not passing the test for potential. Any idea why that is?

ConnorStoneAstro:
> NFW and tNFW not passing the test for potential. Any idea why that is?

Ah, that was because I made the new thin lens default reduced deflection angle compute the gradient of the potential (rather than cyclic definition with physical deflection angle). The way that calculation is done involves a vmap, and NFW/TNFW cant do both vmap and autograd. Nothing changed about NFW/TNFW, they still work, just the test now uses vmap and autograd rather than a very large autograd call.

ConnorStoneAstro:
Posting here the improvements to TNFW. Now works with autodif and vmap (NFW now has this too). Also, better handling of `F` function, is arccosh for `0<x<1-epsilon` and is arccos for `1+epsilon<x`, for `x=1+-epsilon` we use taylor expansion to get correct first and second derivatives. in the images below we compare several approximations where `F_arccosh` is the adopted approximation.
TNFW_F

TNFW_dF

TNFW_d2F

mjyb16 (approved):
Looks like a great improvement! I presume the `#fixme` comment in `_P_tnfw` is fixed now?

PR #344, opened 2025-03-06
perf: faster interp 2d

ConnorStoneAstro:
Refactor for fewer operations


mjyb16 (approved):
Looks good to me, since it passes all the tests (interpolation tests in Caustics are extensive) I think this is ready to go.

PR #346, opened 2025-03-12
feat: adding scale parameters to pixelated models

ConnorStoneAstro:
Sometimes it is nice to have the pixelated model fixed and only change the scaling of the model. For example with a PSF image fitting a star. The scale factor could also be used for a diffusion model operating on images with different units, to keep everything in distribution if the main difference is only a global scaling.


mjyb16 (approved):
Looks good!

PR #350, opened 2025-03-18
update to new list link system

ConnorStoneAstro:
No linker need to call link for all elements of a list for classes like singleplane and multiplane


mjyb16 (approved):
Looks straightforward to me!

PR #354, opened 2025-03-19
feat: adding system to convert between q phi elipses and e1e2 ellipses

ConnorStoneAstro:


mjyb16 (approved):
Bug fixed, now it's good to go!

PR #357, opened 2025-04-02
Performance improvements

rouzib:
# Project Code Changes and Performance Enhancements

List of changes made, with their motivation and performance improvements.

---

## [1]. **Modified chunk_size of lens_source simulator**

**File:** `src/caustics/sims/lens_source.py`  
**Location:** `LensSource.__call__()`  
**Description:**

- Spawning multiple `vmap` for a simple lens is inefficient since the computation is already vectorized.
- Modified `chunk_size` to be a large number (2^24) so that it can be changed if memory issues are experienced, but is
  not used otherwise.
- Using `vmap` without splitting the workload still calls `torch/_functorch/vmap.py(328): _concat_chunked_outputs`.
  Using the `vmap` calls only when necessary (only if `chunk_size` is not `None`) should accelerate `lens_source` by ~
  20%. (Passing `chunk_size=None` to vmap has a 5us overhead, so it might be optimal to keep it like that instead of
  having 4 boolean checks.)

**Performance Impact:**

- Setup: Lens: SIE, Source: Sersic, LensLight: Sersic, 100x100 pixels, quad_level=3 -> ~5x faster

## [2]. **Translate_rotate simultaneous assignment**

**File:** `caustics/src/caustics/utils.py`  
**Location:** `translate_rotate()`  
**Description:**

- Remove unnecessary variable assignment.
- Replaced unnecessary negation and addition by a single subtraction.
- The JIT compiler should do this on its own, so this is of no concern for JITed code.

**Performance Impact:**

- Same setup as previously -> saves ~4ms (initial computation took ~16ms)

## [3]. **k_sersic polynomial efficient computation**

**File:** `caustics/src/caustics/light/func/sersic.py`  
**Location:** `k_sersic()`  
**Description:**

- Apply Horner’s rule for fast polynomial computation.
- Only keep integer divisions and a single tensor division.

**Performance Impact:**

- ~2x faster (~350us vs ~700us)

## [4]. **Simplified sie deflection angles computation**

**File:** `caustics/src/caustics/lenses/func/sie.py`  
**Location:** `reduced_deflection_angle_sie()`  
**Description:**

- Simplify the `q==1.0` numerical instability fix by removing a subtraction.
- Regrouped duplicate computation into intermediary variables.

**Performance Impact:**

- Saves ~500us (from a ~4ms computation)

## [5]. **NFW Helper Methods Optimization**

**File:** `caustics/src/caustics/lenses/func/nfw.py`  
**Location:** `_h_nfw()`, `_g_nfw()`, `_f_nfw()`  
**Description:**

- Simplify control flow of PyTorch where computation by precomputing values.
- Precomputing reduces the overhead of OpenXLA when not using JIT.

**Performance Impact:**

- Saves ~50us (from a ~1.5ms computation)

## [6]. **Removing unnecessary computation in NFW**

**File:** `caustics/src/caustics/lenses/func/nfw.py`  
**Location:** `scale_radius_nfw()`, `scale_density_nfw()`, `_f_nfw()`  
**Description:**

- Removed unnecessary multiplication in `scale_radius_nfw()`.
- Added intermediary variable for `(c + 1)` in `scale_density_nfw()`.

**Performance Impact:**

- Saves ~20us (one elementary operation)

## [7]. **Vectorizing EPL R_Omega**

**File:** `caustics/src/caustics/lenses/func/epl.py`  
**Location:** `_r_omega()`
**Description:**

- Vectorized the function. The function is much faster at the cost of more memory usage. The memory usage scales
  linearly with the number of iterations done, but computation speed scales faster than memory usage.
- Added a `memory_efficient` flag to the function that reverts to the old computation when set to `True`.
- Added a `steps` variable that interpolates between the two implementations by doing the vectorized computation in
  chunks. This implementation trades speed for memory usage.

**Performance Impact:**

- ~6-8x faster when used with 18 iterations to the cost of ~4-5x the memory usage. 

## [8]. **reduced_deflection_angle autograd improvements**

**File:** `caustics/src/caustics/lenses/base.py`  
**Location:** ` reduced_deflection_angle()`  
**Description:**

- Replaced the `torch.func.grad` call by `torch.autograd.functional.vjp` which computes the same gradient, but more efficiently by computing the vector Jacobian product of the function for all pixels at the same time in a backward pass.
- Added the `chunk_size` parameter to let the user change the chunk size. For the new function, a chunk size shouldn't be necessary since its memory usage is about twice the memory usage of the forward pass.

**Performance Impact:**

- Setup: Pixelated convergence with image of 1024x1042 pixels, 1000x1000 meshgrid ->
  - vjp: 1 786 744.99 ± 114 095.83 us, 358.47 MB
  - grad: 15 471 684.12 ± 204 288.37 us, 8679.21 MB (chunk_size=1000 to fit the computation on the GPU)

## [9]. **Use batch in functions called in `pixelated_convergence`**

**File:** `caustics/src/caustics/lenses/pixelated_convergence.py`  
**Location:** `reduced_deflection_angle_pixelated_convergence()`  
**Description:**

- ~~With the use of a batch possible with the new `interp2d`,~~ the `reduced_deflection_angle_pixelated_convergence` was modified to use a batch in the fft, convolution ~~and interpolation~~.


src/caustics/lenses/func/epl.py
```diff
@@ -3,7 +3,7 @@
 from ...utils import translate_rotate, derotate
 
 
-def _r_omega(z, t, q, n_iter):
+def _r_omega(z, t, q, n_iter, n_chunks=1):
```
ConnorStoneAstro, line 6: I believe this function should be removed. The `func._r_omega` is the only one that gets used

src/caustics/lenses/func/pixelated_convergence.py
```diff
@@ -73,6 +73,7 @@ def build_window_pixelated_convergence(window, kernel_shape):
     return torch.clip((1 - r) / window, 0, 1)
 
 
+# @lru_cache(maxsize=4)
```
ConnorStoneAstro, line 76: if this is commented out, should it be removed?

src/caustics/lenses/func/sie.py
```diff
@@ -67,13 +67,19 @@ def reduced_deflection_angle_sie(x0, y0, q, phi, Rein, x, y, s=0.0):
 
     """
     # Handle the case where q = 1.0, numerical instability
-    q = q - torch.where(q == 1.0, 1e-6 * torch.ones_like(q), torch.zeros_like(q))
+    # q = q - torch.where(q == 1.0, 1e-6 * torch.ones_like(q), torch.zeros_like(q))
```
ConnorStoneAstro, line 70: commented line can be removed

src/caustics/light/func/sersic.py
```diff
@@ -22,14 +22,18 @@ def k_sersic(n: Tensor) -> Tensor:
...
-        + 46 / (25515 * n**2)
-        + 131 / (1148175 * n**3)
-        - 2194697 / (30690717750 * n**4)
+        + x
+        * (
+            4 / 405
+            + x * (46 / 25515 + x * (131 / 1148175 - x * (2194697 / 30690717750)))
+        )
     )
+    # return (2 * n - 1 / 3 + 4 / (405 * n) + 46 / (25515 * n ** 2) + 131 / (1148175 * n ** 3) - 2194697 / (
```
ConnorStoneAstro, line 35: can you move this commented code up into the function so we can see its association

src/caustics/lenses/epl.py
```diff
@@ -24,6 +24,8 @@ class EPL(ThinLens):
     ----------
     n_iter: int
         Number of iterations for the iterative solver.
+    n_chunks: int
```
ConnorStoneAstro, line 27: question: is `n_chunks` the most appropriate parameter? All other functions use `chunk_size` and so it might be best to remain consistent

ConnorStoneAstro (changes requested):
Awesome PR! I have some minor comments to address then it should be ready to go. You will also need to address the merge conflicts with the `dev` branch since all PRs must go to dev before going into main

src/caustics/lenses/func/epl.py
rouzib: I believe you are talking about the `_r_omega` in src/caustics/lenses/epl.py instead?

ConnorStoneAstro: ah lol yes

rouzib:
I changed `n_chunks` to `chunk_size` so that it is similar to other parts of the code. I made it so that the fast version (fully parallelized) is the default, i.e. when `chunk_size=None`, then the code interprets it as `chunk_size=n_iter`.

ConnorStoneAstro (approved):
Absolutely fantastic!

PR #370, opened 2025-05-04
fix: work on fixing upsample psf correct npix

### LINKED ISSUE #365: Lens Source when upsampling needs to account for this when determining the FFT size

ConnorStoneAstro:
closes #365

### DIFF
--- a/src/caustics/sims/lens_source.py
+++ b/src/caustics/sims/lens_source.py
@@ -1,5 +1,3 @@
-from copy import copy
-
 from scipy.fft import next_fast_len
 from torch.nn.functional import avg_pool2d, conv2d
 from typing import Optional, Annotated, Literal, Union
@@ -194,6 +192,8 @@ class LensSource(Module):
 
     @upsample_factor.setter
     def upsample_factor(self, value):
+        value = int(value)
+        assert value > 0, f"upsample_factor should be > 0, not {value}"
         self._upsample_factor = value
         self._build_grid()
 
@@ -221,6 +221,10 @@ class LensSource(Module):
 
     @quad_level.setter
     def quad_level(self, value):
+        value = None if value is None else int(value)
+        assert (
+            value is None or value > 0
+        ), f"quad_level should be None or > 0, not {value}"
         self._quad_level = value
         self._build_grid()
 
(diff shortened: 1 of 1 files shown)


src/caustics/sims/lens_source.py
```diff
@@ -248,20 +252,29 @@ def psf_mode(self):
...
+        assert value in (
+            "fft",
+            "conv2d",
+        ), f"psf_mode should be one of 'fft' or 'conv2d', not {value}"
         self._psf_mode = value
         self._build_grid()
 
     def _build_grid(self):
-        self._psf_pad = (self.psf_shape[1] // 2, self.psf_shape[0] // 2)
+        self._psf_pad = (
```
rouzib, line 263: `LensSource` already expects `self.psf` to be upsampled so `self.psf_shape` should already account for this. No need to multiply by `self.upsample_factor` here.

src/caustics/sims/lens_source.py
```diff
@@ -305,9 +317,14 @@ def _unpad_fft(self, x):
         Tensor
             The input tensor without padding.
         """
-        return torch.roll(x, (-self._psf_pad[0], -self._psf_pad[1]), dims=(-2, -1))[
-            ..., : self._s[0], : self._s[1]
-        ]
+        return torch.roll(
```
rouzib, line 320: Same as before, `self._psf_pad` should already account for the upsampling factor.

src/caustics/sims/lens_source.py
rouzib: Without `self._psf_pad` multiplied by `self.upsample_factor` in `_build_grid()` and `_unpad_fft()`

image

With the code as is:

image


* Note, the source, psf and epl parameters used here are all random and `upsample_factor=2`.

rouzib (changes requested):
The psf is multiplied twice by the upsampling factor here since it is already expected to be of the appropriate shape before passing `upsample_factor=n`.

ConnorStoneAstro:
@rouzib thanks for pointing that out! It's what I get for trying to go fast haha. Ok new version and I think I've fixed the stuff you pointed out. I think the thing that was tripping me up was clipping the psf padding after downsampling, so I swapped the order. I tried taking a psf convolved image and changing the upsample factor to test that things looked right and this time it seemed ok, lets see what your test finds!

rouzib (approved):
Everything looks good on my end!

Here are the same tests with `upsample_factor=2`. The now give consistent results and the expected shapes.
image

PR #378, opened 2025-07-31
Fixing requirements.txt

mjyb16:


mjyb16:
> With the new version of Lenstronomy, everything works on my end. Hopefully, future versions won’t break anything.

Great to hear, thanks for checking!

ConnorStoneAstro:
Nice! looks like it works

rouzib:
I did not find any changes that affect Caustics' code in Astropy's changelog so I think we can probably get away with the >6.0.0 (the lower version that still supports python 3.9). However, we might need to force our tests to run with <7.0.0, and this could introduce unchecked bugs introduced in future releases of Astropy. I don't think we have anything to worry about with 7.0.0 and 7.1.0, but our tests won't be able to validate newer versions unless Lenstronomy gets updated.

PR #379, opened 2025-07-31
Warn python version

mjyb16:
Added a warning to the docs and to the ReadME about using Python >=3.12 until we have testing set up for 3.12 (I will then update this warning to >=3.13).

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ Simply install `caustics` from PyPI:
 pip install caustics
 ```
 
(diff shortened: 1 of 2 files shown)


ConnorStoneAstro:
I'm not sure this is the way to go. We can update the `pyproject.toml` file to specify only the python versions that are allowed. But we should probably just try and get version 3.12 to work. If that takes a little while thats ok. I dont think we are losing users quickly by not supporting 3.12

mjyb16:
But I agree, we don't need to merge this to main if 3.12 works off the bat, which it looks like it does.

PR #382, opened 2025-07-31
Upgrade to compatibility with Python 3.12

mjyb16:
Added 3.12 to CI workflow, and I also relaxed the Numba upper version limit. 

More generally, I think that strict version requirements and version upper bounds are generally not best practice for the following reasons:
1. While such reqs aim to improve reproducibility, as a trade-off, they increase the likelihood of conflicts down the road, harming future package maintainers. 
2. Upgrading versions usually fixes bugs rather than introducing new ones. The major exception to this is API changes, but I believe it is the duty of maintainers to adapt to API changes as they occur and set upper version limits as a last resort. 
3. Caustics is less a top-level package, and more of a strong lensing framework which can be built upon by others. Since pinning versions is best done at the top level (see e.g. [the top answer here](https://stackoverflow.com/questions/28509481/should-i-pin-my-python-dependencies-versions)), I believe we should be more cautious about upper bounds, since they are likely to cause cascading version conflicts/increasingly onerous versioning downstream. 

Open to discussing this issue further :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
 
     steps:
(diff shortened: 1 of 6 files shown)


ConnorStoneAstro (approved):
Nice!

PR #432, opened 2026-07-30
Fix multiplane time delay calculations

GuillaumePayeur:
While reviewing multiplane lensing I noticed a few bugs and mistakes. To summarize:

1. An error sometimes occurs when multiplane ray tracing is done with the lenses supplied out of redshift order.
2. The coefficients `tau_ij` have an extra factor of `(1 + z_ls[i])` which is not to be included when using comoving distances (as opposed to angular diameter distances). This affects both the Shapiro and geometric time delays.
3. The geometric time delays are calculated using a factor `(alpha_x**2 + alpha_y**2)` representing the physical deflection angle squared, however the correct factor is the angular position difference squared.

I've corrected these three things while keeping the logic and variable names mostly unchanged. I've also added a few tests in `test_multiplane.py` to test whether, in the case where multiple lenses have the same redshift, combining these lenses into one and using the `SinglePlane` class is equivalent to keeping them separate and using the `Multiplane` class.


src/caustics/lenses/multiplane.py
```diff
@@ -88,36 +88,57 @@ def _raytrace_helper(
...
-                    X * rad_to_arcsec / D_l, Y * rad_to_arcsec / D_l
-                )
-                TD += (-tau_ij * beta_ij * arcsec_to_rad**2) * potential
+                tau_is = D_l * D_s / (D_is * c_Mpc_s) / days_to_seconds
+                potential = self.lenses[i].potential(x_i, y_i)
+                TD += (-tau_is * arcsec_to_rad**2) * potential
+
+            # Coplanar lenses act at the same ray position. Accumulate all their
+            # deflections and potentials before propagating to the next plane.
+            if next_is_coplanar:
```
ConnorStoneAstro: I would prefer not to have value dependent logic like this. It is fatal for vectorization, and almost any use-case with JAX. The extra compute for the null update in the rest of the loop for the coplanar case is worth it to retain vectorizability.

src/caustics/lenses/multiplane.py
```diff
@@ -88,36 +88,57 @@ def _raytrace_helper(
...
+
+            # Propagate rays to the next plane (basically eq 18)
+            D = self.cosmology.transverse_comoving_distance_z1z2(z_l, z_next)
+            D_next = self.cosmology.transverse_comoving_distance(z_next)
+            X_next = X + D * theta_x * arcsec_to_rad
+            Y_next = Y + D * theta_y * arcsec_to_rad
+
             if geometric_time_delay:
-                TD += (tau_ij * arcsec_to_rad**2 * 0.5) * (alpha_x**2 + alpha_y**2)
+                tau_ij = D_l * D_next / (D * c_Mpc_s) / days_to_seconds
```
ConnorStoneAstro: Perhaps use a `where` statement to force tau_ij to zero when D==0?

ConnorStoneAstro (commented):
This is very interesting and you are making me rethink the multiplane implementation. I think the bugs you found are indeed bugs and you've fixed them, very nice! Now with those implementations, are you able to increase the sensitivity in the test against lenstronomy?

Also, would you be interested in writing a second implementation that directly follows equation 16 in McCully 2014? It might just come out cleaner than what I wrote that has now gone through several updates (you can do it all in angular diameter distances if that's easier for you). I notice that even after your updates there are several places where some simplification could be applied (several rad_to_arcsec factors that can be removed at least).

ConnorStoneAstro:
One other thing, could you change the PR to point to dev rather than main?

src/caustics/lenses/multiplane.py
GuillaumePayeur: You're right, thanks for pointing this out. I've made a new commit addressing this

src/caustics/lenses/multiplane.py
GuillaumePayeur: That's what I ended up doing, thanks for the idea

GuillaumePayeur:
I've made a new commit and changed the branch to dev.

Regarding your questions;
> This is very interesting and you are making me rethink the multiplane implementation. I think the bugs you found are indeed bugs and you've fixed them, very nice! Now with those implementations, are you able to increase the sensitivity in the test against lenstronomy?

I think here you're talking about the `test` test in `test_multiplane.py` right? That test uses `lens_test_helper` which does not test time delays, hence my PR doesn't affect it (the only thing my PR does to ray-tracing is avoid an error when redshifts are supplied out of order, but nothing changes quantitatively). So no, I wouldn't be able to increase the sensitivity. It's possible that I misunderstood what you meant though, so let me know :)
 
> Also, would you be interested in writing a second implementation that directly follows equation 16 in McCully 2014? It might just come out cleaner than what I wrote that has now gone through several updates (you can do it all in angular diameter distances if that's easier for you). I notice that even after your updates there are several places where some simplification could be applied (several rad_to_arcsec factors that can be removed at least).

I agree that a new cleaner implementation is possible, but I think that for this PR and for the moment I'd prefer just fixing the bugs, if that's okay with you. I'll consider reaching out to you when I'm less crammed to discuss this.

ConnorStoneAstro:
Hi @GuillaumePayeur, this update is looking great! Could you update your branch to match with `dev`? Then the `CD` test should pass.

Also, could you add time delay comparisons in the multiplane test against lenstronomy? It looks like they have a function for it, so this should be pretty quick for an LLM to wire up.
https://lenstronomy.readthedocs.io/en/latest/lenstronomy.LensModel.MultiPlane.html#lenstronomy.LensModel.MultiPlane.decoupled_multi_plane.MultiPlaneDecoupled.geo_shapiro_delay
Now that you've pointed out multiple bugs I'm a bit paranoid!

GuillaumePayeur:
I remounted the branch from `main` to `dev`, let's see if the `CD` test passes now.

I also added a time delay comparison to lenstronomy in the multiplane case. That additional test is part of the `test_multiplane_time_delay` function in `test_multiplane.py`. And of course the test is passed.

Let me know if there's any other issue!

ConnorStoneAstro (approved):
Amazing job!

